PR #99, opened 2014-02-03
Adding support for custom S3 Object options

meatballhat:
mostly so that I can pass a canned acl of ":public_read", but it felt silly to limit to only one of the [many options available](http://docs.aws.amazon.com/AWSRubySDK/latest/AWS/S3/S3Object.html#write-instance_method).

### DIFF
--- a/lib/dpl/cli.rb
+++ b/lib/dpl/cli.rb
@@ -1,4 +1,5 @@
 require 'json'
+require 'uri'
 require 'dpl/error'
 require 'dpl/provider'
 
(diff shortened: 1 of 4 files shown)


BanzaiMan:
@meatballhat Could you rebase this?

meatballhat:
Oh wow this is ancient.  Sure thing :smiley_cat:

meatballhat:
@BanzaiMan I don't think this is a valid PR anymore.  Handling acl, cache_control, and expires were the big ones, and those are all covered now.

PR #107, opened 2014-03-12
Added additional logging to s3 provider

thetron:
This is a reasonably trivial commit, but I thought it might help to have a bit of logging output for S3 deploys. We've also had some issues on our (really) long S3 deploys hitting the 10 minute timeout, and though the extra logs might help here?

### DIFF
--- a/lib/dpl/provider/s3.rb
+++ b/lib/dpl/provider/s3.rb
@@ -32,8 +32,10 @@ module DPL
       end
 
       def push_app
+        log "Uploading to #{option(:bucket)}"
         Dir.chdir(options.fetch(:local_dir, Dir.pwd)) do
           Dir.glob("**/*") do |filename|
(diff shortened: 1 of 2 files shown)


BanzaiMan:
So sorry for dropping the ball. Could you rebase this and we will take a look soon! Thank you! <3

BanzaiMan:
We've updated S3 provider, so we'll really need a rebase. :-D

thetron:
Sorry @BanzaiMan I totally missed your last comment (in July!). I think - in its current state - the output is verbose enough for me. As much as I'd love to get a contrib in there - I don't really know if there is a real need for this anymore?

BanzaiMan:
@thetron Thanks for the input, and the effort you've put in. I am going to close this for now, then. Thanks again!

PR #201, opened 2014-11-27
Add Chef Supermarket provider

brugidou:
Upload Chef cookbooks to https://supermarket.getchef.com/

#### Options:

 * **user_id**: Required. The user name at Chef Supermarket.
 * **client_key**: Required. The client API key file name.
 * **cookbook_category**: Required. The cookbook category in Supermarket (see: https://docs.getchef.com/knife_cookbook_site.html#id12 )

#### Examples:

    dpl --provider=chef-supermarket --user-id=chef --client-key=.travis/client.pem --cookbook-category=Others

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -59,3 +59,7 @@ group :puppet_forge do
   gem 'puppet'
   gem 'puppet-blacksmith'
 end
+
+group :chef_supermarket do
+  gem 'chef'
+end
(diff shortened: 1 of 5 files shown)


BanzaiMan:
Sorry that it took a while. You can test it with `edge: true`.

brugidou:
I have run a test and there is a dependency problem linked to the ruby version used for dpl. See https://travis-ci.org/criteo-cookbooks/test-cookbook/builds/46628492

I did not know how the chef gem dependency would be added at runtime without bundler, aparently a gem install is done directly and fails with ruby 1.9.3?

PR #210, opened 2014-12-19
respect build exit code in heroku api provider

### LINKED ISSUE #209: heroku api strategy does not detect failure
We had a transient failed push, which ended with the output:

```
=====> Detected Framework: Node.js

 !     Push rejected, failed to compile Multipack app
```

But `dpl` happily exits 0 in this case.  I will research this myself (wanted to get it logged), but there must be a way to check back in on the status of a build once it's finished?

bobzoller:
fixes #209

happy to get feedback as my ruby is rusty and I chose to follow the existing style in this file.  FYI I've tested this manually against a failing build and a successful build.

### DIFF
--- a/lib/dpl/provider/heroku/api.rb
+++ b/lib/dpl/provider/heroku/api.rb
@@ -9,6 +9,7 @@ module DPL
           pack_archive
           upload_archive
           trigger_build
+          verify_build
         end
 
         def archive_file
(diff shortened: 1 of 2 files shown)


BanzaiMan:
@bobzoller Could you add specs, please?

bobzoller:
the existing spec is basically a stub -- is there an example of a great provider spec you can point me at for inspiration?

bobzoller:
@BanzaiMan added specs, feedback welcome.  (my ruby is rusty!)

bobzoller:
I noticed an issue during our production use of the change in this PR -- the latest commit addresses it.

Occasionally, dpl's request to the build result API will beat Heroku to the punch, and the build will still be pending (no exit_code).  In this case, dpl was exiting with an error.  Instead, it should retry and wait for the exit_code to be populated.  I chose to make it an infinite loop for simplicity and because I know travis has other timeout facilities, but I could easily chose an arbitrary number of attempts instead.

I'm testing this in production now, but feedback welcome.

bobzoller:
@BanzaiMan we lived with this for a few days, so I'm confident it works.

BanzaiMan:
@bobzoller Could you try again with `edge: true`? Thank you.

bobzoller:
we're actually using `dpl` outside of Travis, but I did have a successful deploy using `dpl-1.7.10.travis.701.4`.

PR #212, opened 2015-01-07
Dont log the entire API token and add a useful error when no packages ar...

ice799:
...e found

Hi @BanzaiMan just 2 small tweaks to logging I noticed when trying out dpl via a travis-ci build job.


rkh:
Thank you for your contribution.

PR #215, opened 2015-01-13
Only use --upload-dir for pypi docs if docs_dir is set. Fixes travis-ci/...

mikebryant:
travis-ci/travis-ci#2011.

This allows Travis to honor the upload-dir directive in setup.cfg.
This means that standard sphinx docs and pre-generated docs using setup.cfg
will just work out of the box, without special configuration being needed
for Travis.

### DIFF
--- a/lib/dpl/provider/pypi.rb
+++ b/lib/dpl/provider/pypi.rb
@@ -65,7 +65,12 @@ module DPL
       def push_app
         context.shell "python setup.py register -r #{options[:server] || 'pypi'}"
         context.shell "python setup.py #{options[:distributions] || 'sdist'} upload -r #{options[:server] || 'pypi'}"
-        context.shell "python setup.py upload_docs --upload-dir #{options[:docs_dir] || 'build/docs'}"
+        if options[:docs_dir]
+          docs_dir_option = '--upload-dir ' + options[:docs_dir]
+        else
+          docs_dir_option = ''
+        end
+        context.shell "python setup.py upload_docs #{docs_dir_option}"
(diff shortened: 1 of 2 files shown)


Joshua-Anderson:
@mikebryant You broke the tests.

lib/dpl/provider/pypi.rb
```diff
@@ -65,7 +65,7 @@ def needs_key?
       def push_app
         context.shell "python setup.py register -r #{options[:server] || 'pypi'}"
         context.shell "python setup.py #{options[:distributions] || 'sdist'} upload -r #{options[:server] || 'pypi'}"
-        context.shell "python setup.py upload_docs --upload-dir #{options[:docs_dir] || 'build/docs'}"
+        context.shell "python setup.py upload_docs#{options[:docs_dir] ? ' --upload-dir ' + options[:docs_dir] : ''}"
```
BanzaiMan: Do not embed the option logic inside `#{}`, please. This makes the logic unclear.

mikebryant:
@BanzaiMan Is that acceptable?

BanzaiMan:
Use `edge: true` to test it, please.

mikebryant:
Well the automatic running of build_sphinx seems to work: https://travis-ci.org/mikebryant/django-autoconfig/jobs/51875948#L948

Not sure why the authentication is failing though...
I wonder if the upload_docs comand needs the -r #{options[:server] || 'pypi'} like the previous commands do - maybe it's never worked as it currently is?

PR #228, opened 2015-03-06
Update deis provider to allow for custom scheme

jonpaul:
Deis client won't follow redirects. This is a problem if you have set up deis to force an https scheme. This allows you to optionally provide a controller scheme. Really trivial change, made the edit live on GitHub so, no tests ran.

### DIFF
--- a/lib/dpl/provider/deis.rb
+++ b/lib/dpl/provider/deis.rb
@@ -101,7 +101,11 @@ module DPL
       end
 
       def controller_url
-        "http://#{option(:controller)}"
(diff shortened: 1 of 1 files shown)


jonpaul:
Looks like this is green, please help me understand any changes necessary to incorporate this trivial but useful change.

jonpaul:
This actually doesn't work at all. I did manage to write my own Deis DPL that is working just fine for us, but makes significant changes. PIP packages of deis CLI are awful and should never be used, as an aside.

PR #230, opened 2015-03-10
Specifying s3 folder location for beanstalk deploy

### LINKED ISSUE #172: elastic beanstalk - specify folder to deploy
Is it possible to specify which folder to deploy to elastic beanstalk on a build.
Similar to **local-dir**  for s3 deployments. Similarly, i would like travis to not clean up built files (as these are the ones i want to deploy)  s3 deploy's achieve this through **skip_cleanup**

zabawaba99:
This is an attempt to resolve #172.

I'd like to add some tests for this but I'm not too sure how to go about testing it. Any suggestions?

### DIFF
--- a/lib/dpl/provider/elastic_beanstalk.rb
+++ b/lib/dpl/provider/elastic_beanstalk.rb
@@ -54,10 +54,6 @@ module DPL
         option(:bucket_name)
       end
 
-      def bucket_path
-        options[:bucket_path]
-      end
-
       def s3
         @s3 ||= AWS::S3.new
       end
@@ -93,12 +89,11 @@ module DPL
 
       def upload(key, file)
         obj = s3.buckets[bucket_name]
-        file_path = bucket_path
-        if file_path
-          file_path << '/' unless file_path.end_with?('/')
+        if option(:bucket_path)
+          option(:bucket_path).gsub!(/\/*$/,'/')
         end
 
-        obj = obj.objects["#{file_path}#{key}"]
+        obj = obj.objects["#{option(:bucket_path)}#{key}"]
         obj.write(Pathname.new(file))
         obj
       end
--- a/lib/dpl/provider/elastic_beanstalk.rb
+++ b/lib/dpl/provider/elastic_beanstalk.rb
@@ -54,6 +54,10 @@ module DPL
         option(:bucket_name)
       end
 
+      def bucket_path
+        option(:bucket_path)
+      end
+
       def s3
         @s3 ||= AWS::S3.new
       end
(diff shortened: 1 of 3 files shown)


lipis:
Would be nice to have that merged if it's working :+1:

zabawaba99:
Is there anything else that this PR needs to get merged in?

lib/dpl/provider/elastic_beanstalk.rb
```diff
@@ -88,7 +92,13 @@ def create_zip
       end
 
       def upload(key, file)
-        obj = s3.buckets[bucket_name].objects[key]
+        obj = s3.buckets[bucket_name]
+        file_path = bucket_path
+        if file_path
+          file_path << '/' unless file_path.end_with?('/')
```
BanzaiMan: This combination of `if` and `unless` doesn't read well.

`file_path.gsub!(/\/*$/,'/')` should be sufficient.

Also, what is the reason for `file_path = bucket_path`? I don't think `bucket_path` is used elsewhere, and just having `file_path` pointing to the same object as `bucket_path` does not make sense to me.

spec/provider/elastic_beanstalk_spec.rb
```diff
@@ -15,11 +15,12 @@
   let(:app) { 'example-app' }
   let(:env) { 'live' }
   let(:bucket_name) { "travis-elasticbeanstalk-test-builds-#{region}" }
+  let(:bucket_path) { "some/app"}
 
   subject :provider do
     described_class.new(
       DummyContext.new, :access_key_id => access_key_id, :secret_access_key => secret_access_key,
-      :region => region, :app => app, :env => env, :bucket_name => bucket_name
+      :region => region, :app => app, :env => env, :bucket_name => bucket_name, :bucket_path => bucket_path
```
BanzaiMan: Does this actually add any test to the change proposed in `#upload`?

lib/dpl/provider/elastic_beanstalk.rb
zabawaba99: > This combination of if and unless doesn't read well.
>
> `file_path.gsub!(/\/*$/,'/')` should be sufficient.

Fair point. I'll change it up.

> Also, what is the reason for file_path = bucket_path? I don't think bucket_path is used elsewhere, and just having file_path pointing to the same object as bucket_path does not make sense to me.

The reason that I created a `bucket_path` variable was to attempt to stay in line with the provider itself. I noticed that there were methods created for every option that could be passed in.

I could do something like

```ruby
path = options[:bucket_path]
if path
    path.gsub!(/\/*$/,'/')
end

obj = obj.objects["#{path}#{key}"]
obj.write(Pathname.new(file))
obj
```

instead. Or were you talking about modifying the value inside of the `options` hash?

spec/provider/elastic_beanstalk_spec.rb
zabawaba99: No. I had made a comment on the PR itself that I wasn't too sure on how to actually go about testing this. I tried digging around through the code to use other providers as examples but came up dry.

zabawaba99:
@BanzaiMan d4786c6 cleans up the code a bit and adds some tests

PR #232, opened 2015-03-13
release_number, not release-number

neoascetic:


Joshua-Anderson:
Doesn't matter, they bdoth should work.

neoascetic:
Hmm, I didn't know. 
But mixing of underscore-separated and dash-separated options in one block looks a little bit ugly, doesn't it?

Joshua-Anderson:
:smile: That's what happens when a bunch of people edit e same file. There are more `-` in the options block than `_` though, but I don't ccare much either way.

neoascetic:
@Joshua-Anderson yeah, didn't notice that :D

PR #235, opened 2015-03-19
Skip installing a gem if it's already installed

BanzaiMan:
In particular, we should skip 'gem install nokogiri' if
it's already available.


svenfuchs:
WHOOP!

PR #238, opened 2015-03-23
avoid false positivies in the case of aws-sdk-v1 and aws-sdk

iamruinous:


lib/dpl/provider.rb
```diff
@@ -59,7 +59,7 @@ def self.requires(name, options = {})
       load    = options[:load]    || name
       gem(name, version)
     rescue LoadError
-      context.shell("(gem list %s | grep -q -w %s) || gem install %s -v %p --no-ri --no-rdoc #{'--pre' if options[:pre]}" % [name, name, name, version], retry: true)
+      context.shell("(gem list %s | grep -q -w %s[[:space:]]) || gem install %s -v %p --no-ri --no-rdoc #{'--pre' if options[:pre]}" % [name, name, name, version], retry: true)
```
BanzaiMan: I'm guessing that we need a `[[:space:]]` before the gem name, too.

iamruinous:
Updated to check beginning of line as well

BanzaiMan:
@iamruinous Thanks for the PR, and I apologize for the issue you had. We've decided to revert this check (which was—perhaps incompletely–tested) for the 1.7.11 release.

We will revisit this issue for a future release.

Thanks again!

PR #257, opened 2015-05-07
Make bucket_path option truly optional for elastic beanstalk provider

JamesAwesome:
PR #230 relied on the `#option` method in order to access the value of the `bucket_path` option. Because of this if :bucket_path was left empty or nil the elastic beanstalk provider would error with `Missing bucket_path`. 

Instead add an accessor method for `bucket_path` which allows for :bucket_path to be nil this way if `bucket_path` is not specified deploys still work.

### DIFF
--- a/lib/dpl/provider/elastic_beanstalk.rb
+++ b/lib/dpl/provider/elastic_beanstalk.rb
@@ -95,11 +95,14 @@ module DPL
 
       def upload(key, file)
         obj = s3.buckets[bucket_name]
+
         if option(:bucket_path)
           option(:bucket_path).gsub!(/\/*$/,'/')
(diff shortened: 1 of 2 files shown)


zabawaba99:
Sorry about that @JamesAwesome . I went back and forth a few times on that implementation and thought I was accessing the hash directly. :+1:

JamesAwesome:
Figured as much :smile:

BanzaiMan:
Thanks for the PR. Is there a spec to test this?

JamesAwesome:
Added spec to test a provider without the bucket_path option specified calls #push_app properly.

JamesAwesome:
:metal: Thanks!

PR #256, opened 2015-05-03
Cleaned up bintray provider pull request

rtyler:
This is just a bit of rebasing and tidying up of #255, which can be closed now.

It contains only @eyalbe4's original changes, I just rebased the commits together a bit and updated some commit messages.

PR #259, opened 2015-05-17
Bintray provider

eyalbe4:
This pull request is following https://github.com/travis-ci/dpl/pull/256
It includes all comments provided by @rtyler and @BanzaiMan
Please let me know if anything else is needed to add the new Bintray provider and thanks a lot for helping with this!

### DIFF
--- a/lib/dpl/provider/bintray.rb
+++ b/lib/dpl/provider/bintray.rb
@@ -6,8 +6,6 @@ require 'find'
 module DPL
   class Provider
     class Bintray < Provider
-      @test_mode = false
-
       def check_auth
       end
 
@@ -15,9 +13,18 @@ module DPL
         false
       end
 
-      attr_writer :test_mode
+      attr_accessor :test_mode
+      attr_reader :user
+      attr_reader :key
+      attr_reader :file
+      attr_reader :passphrase
+      attr_reader :url
+      attr_reader :dry_run
+      attr_reader :descriptor
 
-      def init_from_args
+      def initialize(*args)
+        super(*args)
+        @test_mode = false
         @user = options[:user]
         @key = options[:key]
         @url = options[:url]
@@ -43,8 +50,8 @@ module DPL
       end
 
       def read_descriptor
-        log "Reading descriptor file: #{@file}"
-        @descriptor = JSON.parse(File.read(@file))
+        log "Reading descriptor file: #{file}"
+        @descriptor = JSON.parse(File.read(file))
       end
 
       def descriptor=(json)
@@ -52,9 +59,9 @@ module DPL
       end
 
       def head_request(path)
-        url = URI.parse(@url)
+        url = URI.parse(self.url)
         req = Net::HTTP::Head.new(path)
-        req.basic_auth @user, @key
+        req.basic_auth user, key
 
         sock = Net::HTTP.new(url.host, url.port)
         sock.use_ssl = true
@@ -66,12 +73,12 @@ module DPL
       def post_request(path, body)
         req = Net::HTTP::Post.new(path)
         req.add_field('Content-Type', 'application/json')
-        req.basic_auth @user, @key
+        req.basic_auth user, key
         if !body.nil?
           req.body = body.to_json
         end
 
-        url = URI.parse(@url)
+        url = URI.parse(self.url)
         sock = Net::HTTP.new(url.host, url.port)
         sock.use_ssl = true
         res = sock.start {|http| http.request(req) }
@@ -79,7 +86,7 @@ module DPL
       end
 
       def put_file_request(local_file_path, upload_path, matrix_params)
-        url = URI.parse(@url)
+        url = URI.parse(self.url)
 
         data = File.read(local_file_path)
         http = Net::HTTP.new(url.host, url.port)
@@ -94,7 +101,7 @@ module DPL
         end
 
         request = Net::HTTP::Put.new("#{upload_path}")
-        request.basic_auth @user, @key
+        request.basic_auth user, key
         request.body = data
 
         return http.request(request)
@@ -103,22 +110,24 @@ module DPL
       def upload_file(artifact)
         log "Uploading file '#{artifact.local_path}' to #{artifact.upload_path}"
 
-        package = @descriptor["package"]
-        version = @descriptor["version"]
+        if dry_run
+          return
+        end
+
+        package = descriptor["package"]
+        version = descriptor["version"]
         package_name = package["name"]
         subject = package["subject"]
         repo = package["repo"]
         version_name = version["name"]
 
         path = "/content/#{subject}/#{repo}/#{package_name}/#{version_name}/#{artifact.upload_path}"
-        if !@dry_run
-          res = put_file_request(artifact.local_path, path, artifact.matrix_params)
-          log_bintray_response(res)
-        end
+        res = put_file_request(artifact.local_path, path, artifact.matrix_params)
+        log_bintray_response(res)
       end
 
       def package_exists_path
-        package = @descriptor["package"]
+        package = descriptor["package"]
         subject = package["subject"]
         name = package["name"]
         repo = package["repo"]
@@ -127,7 +136,7 @@ module DPL
 
       def package_exists?
         path = package_exists_path
-        if !@dry_run
+        if !dry_run
           res = head_request(path)
           code = res.code.to_i
         else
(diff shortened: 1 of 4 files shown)


eyalbe4:
Hi. Did you get a chance to look at our code?

lib/dpl/provider/bintray.rb
```diff
@@ -0,0 +1,532 @@
...
+
+      def init_from_args
+        @user = options[:user]
+        @key = options[:key]
+        @url = options[:url]
+        @file = options[:file]
+        @passphrase = options[:passphrase]
+        @dry_run = options[:dry_run]
+
+        if @user.nil?
```
BanzaiMan: I *think* `@user` can get the value `false` in this context, as unlikely as it might be.

lib/dpl/provider/bintray.rb
```diff
@@ -0,0 +1,532 @@
...
+        end
+        if @url.nil?
+          @url = 'https://api.bintray.com'
+        end
+        if @dry_run.nil?
+          @dry_run = false
+        end
+      end
+
+      def get_user
```
BanzaiMan: In Ruby, I suggest not using `get_*` and `set_*` as method names.

Here, we can just write:

```ruby
attr_reader :user
```
and use `user` to get the value `@user`.

lib/dpl/provider/bintray.rb
```diff
@@ -0,0 +1,532 @@
...
+        @user
+      end
+
+      def read_descriptor
+        log "Reading descriptor file: #{@file}"
+        @descriptor = JSON.parse(File.read(@file))
+      end
+
+      def set_descriptor(json)
```
BanzaiMan: A better method name is `descriptor=`.

lib/dpl/provider/bintray.rb
eyalbe4: Please let me know if I'm missing something.
I did the following test:
```
@user = options[:user]
str = "false"
bool = false

if str == @user
   puts '1'
end
if bool == @user
   puts '2'
end
```
The output I git is 1.
What do you think? Can we leave the code as is?

lib/dpl/provider/bintray.rb
eyalbe4: Got it. Thanks. I noticed though that this method is not needed, so I removed it.
I implemented this in other places.

lib/dpl/provider/bintray.rb
eyalbe4: Done.

eyalbe4:
Hi @BanzaiMan , Please let me know if you agree with my reply to your first comment or you'd like me to change the code.
I implemented the rest of your comments.
Thanks!

lib/dpl/provider/bintray.rb
BanzaiMan:
```yaml
user: false
```
I believe will result in `abort`, I believe.

```
$ pry                
[1] pry(main)> require 'yaml'
=> true
[2] pry(main)> YAML.load "foo: false"
=> {"foo"=>false}
```

lib/dpl/provider/bintray.rb
```diff
@@ -0,0 +1,515 @@
...
+    class Bintray < Provider
+
+      @file
+      @user
+      @key
+      @passphrase
+      @url
+      @descriptor
+      @dry_run
+      @test_mode = false
```
BanzaiMan: In Ruby, you don't have to declare instance variables to us them.

lib/dpl/provider/bintray.rb
```diff
@@ -0,0 +1,515 @@
...
+
+      def check_auth
+      end
+
+      def needs_key?
+        false
+      end
+
+      def set_test_mode
```
BanzaiMan: Did I mention elsewhere that `set_*` and `get_*` method names are not very Ruby-like?

The method should be `test_mode=(bool)`, and not necessarily assume it's going to set to `true` and never to `false` (or whatever it might be).

For example:

```ruby
def test_mode=(bool)
  @test_mode = bool
end
```

In general, though, if the instance variable name and the method names are "similar", you can do away with:

```ruby
attr_writer :test_mode
```

lib/dpl/provider/bintray.rb
eyalbe4: Done.

lib/dpl/provider/bintray.rb
eyalbe4: Done.

eyalbe4:
Thanks for your comments @BanzaiMan 
Both have been implemented.

eyalbe4:
Can this PR be merged?
Please let us know if anything else is required from us.
Thanks!

yoav:
Hi, is there anything else we (jfrog) can do to/need to do to have this pull request merged?

Thanks

joshk:
Sorry for the delay on this, we will get to it shortly once we have finished some improvements to our caching support. Thank you for the awesome contribution.

yoav:
@joshk No worries! Thanks.

lib/dpl/provider/bintray.rb
```diff
@@ -0,0 +1,505 @@
...
+      def check_auth
+      end
+
+      def needs_key?
+        false
+      end
+
+      attr_writer :test_mode
+
+      def init_from_args
```
BanzaiMan: I'm a bit confused by this method. Why is the initialization code not in `#initialize`?

lib/dpl/provider/bintray.rb
```diff
@@ -0,0 +1,505 @@
...
+          @url = 'https://api.bintray.com'
+        end
+        if @dry_run.nil?
+          @dry_run = false
+        end
+      end
+
+      def read_descriptor
+        log "Reading descriptor file: #{@file}"
+        @descriptor = JSON.parse(File.read(@file))
```
BanzaiMan: Using instance variables everywhere is generally not a good idea. Either use the attribute accessor (or reader), so that if one ever wants to change the way it is accessed it can be done in just one place.

lib/dpl/provider/bintray.rb
```diff
@@ -0,0 +1,505 @@
...
+
+        package = @descriptor["package"]
+        version = @descriptor["version"]
+        package_name = package["name"]
+        subject = package["subject"]
+        repo = package["repo"]
+        version_name = version["name"]
+
+        path = "/content/#{subject}/#{repo}/#{package_name}/#{version_name}/#{artifact.upload_path}"
+        if !@dry_run
```
BanzaiMan: This is a bit strange place to check the value of `@dry_run`. There is nothing this method should do if `@dry_run` is truth-y, so it should be checked up front:

```ruby
def upload_file(artifact)
  return unless @dry_run
  ⋮
end
```

(Having a method named `dry_run?` would be more Ruby-like.)

lib/dpl/provider/bintray.rb
eyalbe4: Moved the code to #initialize as requested.

lib/dpl/provider/bintray.rb
eyalbe4: Done. 
I'm now referencing the instance variables only from inside the #initialize methods and the #read_descriptor method, to protect those variables from write access (from outside the class).
If you'd like me to add accessors to all of them and reference them through the accessors from within the above methods as well, please let me know.

lib/dpl/provider/bintray.rb
eyalbe4: Done. I still need the log message to be displayed in dry-run mode, so I placed the dry-run check right after the log message.

eyalbe4:
Thank you @BanzaiMan !

BanzaiMan:
You can use this with

```yaml
deploy:
  provider: bintray
  edge: true
  ⋮
```

Once we have a release, you can drop `edge: true`.

andyli:
Would be good to document this on http://docs.travis-ci.com/user/deployment/

eyalbe4:
Sure.
How can we add the new Bintray Provider added to this documentation page?

andyli:
I believe you can do that with a PR to https://github.com/travis-ci/docs-travis-ci-com

eyalbe4:
Thanks @andyli . Pull request submitted:
https://github.com/travis-ci/docs-travis-ci-com/pull/300

hectcastro:
Best I can tell, these changes don't include support for publishing to Maven repositories. Is that accurate?

eyalbe4:
@hectcastro , the tool supports upload to all repository types on Bintray, including Maven repositories.

PR #262, opened 2015-05-22
Allow include_git option to include .git in zip file for elasticbeanstalk deployment

ianblenke:


BanzaiMan:
Thanks for the PR. Could you elaborate on why this is a good feature?

ianblenke:
The primary purpose of this would be to support submodules.

We ended up solving this with a .dockerignore file for awsebcli to include the .git directory. This PR is no longer necessary for us.

BanzaiMan:
OK. I am going to close this for now. We will revisit if we have other requests for the use case described.

PR #264, opened 2015-05-27
Environment variables for Elastic Beanstalk

jasny:
* Allow setting the environment name using environment variable `ELASTIC_BEANSTALK_ENV`. Use case: set the env name based on the current branch.
* Allow setting the version label for elastic beanstalk through environment variable `ELASTIC_BEANSTALK_LABEL`. Falls back to travis sha.


BanzaiMan:
Seems reasonable. Could you add documentation and specs, please?

PR #265, opened 2015-05-27
Use environment variables AWS_ACCESS_KEY_ID and AWS_SECRET_ACCESS_KEY

jasny:
Try the env variables for AWS authentication if these options aren't set in the configuration.
For CodeDeploy, S3, OpsWorks and ElasticBeanstalk.

These environment variables are also [used by the aws cli app](http://docs.aws.amazon.com/cli/latest/userguide/cli-chap-getting-started.html#config-settings-and-precedence).

This allows you to define the credentials once in the configuration, while deploying to multiple AWS services and/or using the AWS CLI.

### DIFF
--- a/README.md
+++ b/README.md
@@ -213,6 +213,11 @@ As a rule of thumb, you should switch to the Git strategy if you run into issues
 
 It is possible to set file-specific `Cache-Control` and `Expires` headers using `value: file[, file]` format.
 
+#### Environment variables:
+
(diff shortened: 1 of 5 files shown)


BanzaiMan:
Sounds reasonable. (I'm a bit surprised we didn't have this already. Oops!) Please add documentation and specs. Thank you!!

PR #269, opened 2015-06-02
CodeDeploy provider support for --waint-untill-deployed option

montaro:


lib/dpl/provider/code_deploy.rb
```diff
@@ -53,17 +53,38 @@ def github_revision
...
+          log "Deployment successful."
+        else
+          error "Deployment failed."
+        end
       rescue Aws::CodeDeploy::Errors::DeploymentLimitExceededException => exception
         error(exception.message)
       end
 
+      def wait_until_deployed(deployment_id)
+        deployment = nil
```
BanzaiMan: Is this necessary?

lib/dpl/provider/code_deploy.rb
```diff
@@ -53,17 +53,38 @@ def github_revision
...
+        loop do
+          result = code_deploy.get_deployment(deployment_id: deployment_id)
+          deployment = result[:deployment_info]
```
BanzaiMan: Do we know `result` would respond to `[]`?

lib/dpl/provider/code_deploy.rb
```diff
@@ -53,17 +53,38 @@ def github_revision
...
+          break unless deployment[:status] == "Created" || deployment[:status] == "Queued" || deployment[:status] == "InProgress"
```
BanzaiMan: We'll have a `NoMethodError` here if `deployment` doesn't respond to `[]` (in particular, if it's `nil`).

BanzaiMan: Are these the only final state we should expect? Is there any error condition we should notice and inform the users about?

BanzaiMan: If `Succeeded` is the only status we should care about, then maybe the `break` statement should be

```ruby
    break if deployment[:status] == 'Succeeded"
```

instead.

lib/dpl/provider/code_deploy.rb
kfateem: I believe its needed. A NameError will be thrown on line 85 if not initialized.

BanzaiMan: Ah, OK. I understand now.

lib/dpl/provider/code_deploy.rb
kfateem: result responds to []. I've included relevant snippets from the [aws ruby sdk api](https://github.com/aws/aws-sdk-ruby/blob/master/aws-sdk-core/apis/codedeploy/2014-10-06/api-2.json) for GetDeployment:

 "GetDeployment":{
      "name":"GetDeployment",
      "http":{
        "method":"POST",
        "requestUri":"/"
      },
      "input":{"shape":"GetDeploymentInput"},
      "output":{"shape":"GetDeploymentOutput"},
      "errors":[
        {
          "shape":"DeploymentIdRequiredException",
          "exception":true
        },
        {
          "shape":"InvalidDeploymentIdException",
          "exception":true
        },
        {
          "shape":"DeploymentDoesNotExistException",
          "exception":true
        }
      ]
    }

"GetDeploymentOutput":{
      "type":"structure",
      "members":{
        "deploymentInfo":{"shape":"DeploymentInfo"}
      }
    }

lib/dpl/provider/code_deploy.rb
BanzaiMan: You'd probably want

```ruby
deployment = {}
```

which agrees with the general type that `loop` is expected to assign to `deployment`. (I understand that the value is overwritten, but it reads better to me.)

lib/dpl/provider/code_deploy.rb
kfateem: deployment should always respond to [ ].

result[:deployment_info] will always return a structure.

kfateem: I believe the error states are the 3 exceptions on GetDeployment:
```json
"errors":[
        {
          "shape":"DeploymentIdRequiredException",
          "exception":true
        },
        {
          "shape":"InvalidDeploymentIdException",
          "exception":true
        },
        {
          "shape":"DeploymentDoesNotExistException",
          "exception":true
        }
```

Currently the errors are re-thrown.

kfateem: The possible status values are:

```json
"DeploymentStatus":{
      "type":"string",
      "enum":[
        "Created",
        "Queued",
        "InProgress",
        "Succeeded",
        "Failed",
        "Stopped"
      ]
    }
```
The deployment should wait if status is "Created",  "Queued", or "InProgress".
Lines 67-71 will test for success:
```ruby
  if deployment[:status] == 'Succeeded'
+          log "Deployment successful."
+        else
+          error "Deployment failed."
+        end
```
Is this sufficient?

lib/dpl/provider/code_deploy.rb
kfateem: Makes sense. Will update.

lib/dpl/provider/code_deploy.rb
BanzaiMan: Ah, I see.

Logically, though, it seems more natural to me to "break out of the `loop` when the deployment reached its final state (`Succeeded`, `Failed`, or `Stopped`)". The current logic reads "do not break if the state is one of these", which requires an extra step in my mind.

The handling of the deployment status in `push_app` seems OK.

kfateem: Good point. Adjust loop to break on final state for better readability.

BanzaiMan:
Once https://travis-ci.org/travis-ci/dpl/builds/65953056 passes, you can test it with:

```yaml
deploy:
  provider: code_deploy
  ⋮
  edge: true
```

PR #279, opened 2015-06-29
Update README.md

angeloxlr8:
Ninefold.com no longer provide a Rails platform so these instructions are no longer applicable.


BanzaiMan:
Thanks.

Issue #283, opened 2015-07-13
Not supported lambda.upload_function


k-kinzal:
```
/home/travis/.rvm/gems/ruby-1.9.3-p551/gems/dpl-1.7.12/lib/dpl/provider/lambda.rb:23:in `push_app': undefined method `upload_function' for #<Aws::Lambda::Client> (NoMethodError)
	from /home/travis/.rvm/gems/ruby-1.9.3-p551/gems/dpl-1.7.12/lib/dpl/provider.rb:137:in `block in deploy'
	from /home/travis/.rvm/gems/ruby-1.9.3-p551/gems/dpl-1.7.12/lib/dpl/cli.rb:41:in `fold'
	from /home/travis/.rvm/gems/ruby-1.9.3-p551/gems/dpl-1.7.12/lib/dpl/provider.rb:137:in `deploy'
	from /home/travis/.rvm/gems/ruby-1.9.3-p551/gems/dpl-1.7.12/lib/dpl/cli.rb:32:in `run'
	from /home/travis/.rvm/gems/ruby-1.9.3-p551/gems/dpl-1.7.12/lib/dpl/cli.rb:7:in `run'
	from /home/travis/.rvm/gems/ruby-1.9.3-p551/gems/dpl-1.7.12/bin/dpl:5:in `<top (required)>'
	from /home/travis/.rvm/gems/ruby-1.9.3-p551/bin/dpl:23:in `load'
	from /home/travis/.rvm/gems/ruby-1.9.3-p551/bin/dpl:23:in `<main>'
```
Error occurs when using the lambda provider.

https://github.com/travis-ci/dpl/blob/master/lib/dpl/provider/lambda.rb#L23
Using the `lambda.upload_function`, but it does not already support .

http://docs.aws.amazon.com/lambda/latest/dg/API_CreateFunction.html
http://docs.aws.amazon.com/lambda/latest/dg/API_UpdateFunctionCode.html
http://docs.aws.amazon.com/lambda/latest/dg/API_UpdateFunctionConfiguration.html
Now it has been to use the `lambda.create_function` or `lambda.update_function_code` + `lambda.update_function_configuration`.

BanzaiMan:
Do you have a build log URL that shows this problem?

k-kinzal:
https://travis-ci.org/k-kinzal/lambda-sns-to-chatwork-function#L158
It will be here.

BanzaiMan:
[...]

PR #284, opened 2015-07-14
Use specific version of aws-sdk for the Lambda provider.

wtcross:
This commit makes the lambda provider temporarily depend
on 2.0.37. This is because of the fact that when the
provider was implemented Lambda was in preview (version
2014-11-11). Now the Lambda API is at version 2015-03-31.

The new API is different than the old one, which breaks
the provider for this project.

This is a fix so that the existing provider works, but
the provider needs to be updated to conform to the newest
API.

### DIFF
--- a/lib/dpl/provider/lambda.rb
+++ b/lib/dpl/provider/lambda.rb
@@ -5,7 +5,7 @@ require 'fileutils'
 module DPL
   class Provider
     class Lambda < Provider
-      requires 'aws-sdk', pre: true
+      requires 'aws-sdk', version: '2.0.37'
       requires 'rubyzip', load: 'zip'
 
       def lambda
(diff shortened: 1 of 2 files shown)


wtcross:
This PR addresses #283

wtcross:
[This](https://github.com/aws/aws-sdk-ruby/tree/v2.0.38/aws-sdk-core/apis/lambda) tag of `aws-sdk` introduces the new API version. The fix in this PR should be considered just a fix. I'm going to create another PR soon that actually updates the lambda provider to use the new API.

wtcross:
@BanzaiMan Are you okay with considering this a fix? I think it makes sense because the current lambda provider implementation is valid against `2014-11-11` version of the Lambda API.

I do plan on updating the provider, to be clear :)

wtcross:
[Here is a test build](https://travis-ci.org/wtcross/lambda-provider-test/builds/70894780) that worked with these changes.

Issue #253, opened 2015-04-27
PyPI provider should use twine for registering and upload


jezdez:
Twine is a tool from the pip makers, see https://pypi.python.org/pypi/twine for why you should use it instead of `python setup.py register` and `python setup.py upload`.

alex:
Without this fixed, passwords are being transmitted in cleartext, in addition to allowing the upload itself to be MITM.

jezdez:
Ah, good point, I should have mentioned that. Some more info: https://packaging.python.org/en/latest/distributing.html#upload-your-distributions

BanzaiMan:
Could anyone take a look at improving https://github.com/travis-ci/dpl/blob/56ab84ab9ccb901a8fedc2bfee6a91c8ab7dc0e1/lib/dpl/provider/pypi.rb?

jezdez:
@alex @dstufft Mind taking a look at #288?

jezdez:
#288 continues to use the setup.py register command which has it's own [issues](https://packaging.python.org/en/latest/distributing.html#register-your-project) but at least this should transmit the registration form data via TLS.

alex:
@BanzaiMan this can be closed now :-)

d0ugal:
Awesome. Thanks @jezdez!

BanzaiMan:
I'll need to cut a dpl release before everyone can use it. :-D

BanzaiMan:
That is, right now, it requires `edge: true`.

BanzaiMan:
OK. dpl 1.7.17 is out with this.

jezdez:
@BanzaiMan @alex I'm afraid that #295 is needed as well to make this bullet proof.

Also could you point me to the repo where the Travis-CI service docs are hosted so I can add a note about the fact that this change would also require to do a project registration on https://pypi.python.org/ once when using the PyPI feature?

dstufft:
FWIW that documentation will only be required until PyPI 2.0 is done, at that point simply uploading a project for the first time will also register it.

BanzaiMan:
1.7.18 is out as well.

BanzaiMan:
@jezdez @dstufft Why can't twine be user-installed? (See the issue above)

dstufft:
Because ``--user`` doesn't work inside of a virtual environment.

PR #288, opened 2015-07-20
Use twine for PyPI uploads.

jezdez:
Fix for https://github.com/travis-ci/dpl/issues/253

### DIFF
--- a/lib/dpl/provider/pypi.rb
+++ b/lib/dpl/provider/pypi.rb
@@ -11,6 +11,7 @@ module DPL
 
       def initialize(*args)
         super(*args)
+        self.class.pip 'twine'
         self.class.pip 'wheel' if options[:distributions].to_s.include? 'bdist_wheel'
       end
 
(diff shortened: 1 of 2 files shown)


alex:
looks reasonable to me!

lib/dpl/provider/pypi.rb
```diff
@@ -64,7 +65,9 @@ def needs_key?
 
       def push_app
         context.shell "python setup.py register -r #{options[:server] || 'pypi'}"
```
dstufft: You might want to remove this line? Unless the project in question has never been registered with PyPI *ever* in it's entire history this line is not important and will negate any security benefits of using twine.

dstufft:
One comment, otherwise looks good to me.

lib/dpl/provider/pypi.rb
dstufft: I should mention too, that Warehouse removes even the "you must have ran ``setup.py register`` at least once ever" caveat too.

jezdez: I think that implies that the docs need to be updated that people have to register their project manually on the PyPI website to be safe?

jezdez: The Travis docs, that is.

dstufft: Yea, quite possibly. Not sure I ever read the docs but currently the only safe way is to do a one time register via the website (can just use version 0 and delete it) and then just use ``twine upload`` in the future.

PR #292, opened 2015-07-30
[packagecloud] Allow users to specify a package_glob we will search.

kylev:
In some cases, eligible package or package-like objects may be laying
around (particularly .gem files in .bundle/ from a bundler driven test
suite). This allows the user to specify the glob(s) for the paths we
will search rather than a recursive search of everything.

This adds the option package_glob to the provider. It takes a string
or list of strings. It defaults to "**/*" (search everthing for package
files).

### DIFF
--- a/lib/dpl/provider/packagecloud.rb
+++ b/lib/dpl/provider/packagecloud.rb
@@ -86,7 +86,7 @@ module DPL
 
       def push_app
         packages = []
-        glob_args = ["**/*"]
+        glob_args = Array(options.fetch(:package_glob, '**/*'))
         Dir.chdir(options.fetch(:local_dir, Dir.pwd)) do
           Dir.glob(*glob_args) do |filename|
             unless File.directory?(filename)
--- a/spec/provider/packagecloud_spec.rb
+++ b/spec/provider/packagecloud_spec.rb
@@ -32,4 +32,25 @@ describe DPL::Provider::Packagecloud do
 
   end
 
+  describe "#push_app" do
+    it 'defaults to searching everywhere' do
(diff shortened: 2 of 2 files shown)


kylev:
To reproduce:
- Create a trivial gem with a gemspec containing a few dependencies.
- Add a Gemfile that refers to `gemspec`.
- Set up basic ruby .travis.yml with a `packagecloud` deploy section.
- Set `cache: bundler`.
- Add a `gem build wahtever.gemspec` script step.

Because of the Gemfile, bundle install will be run. Because of caching, all gems and dependencies will be placed in `vendor/`. Because of the `**/*` glob, all those vendor .gem files will be pushed to your PackageCloud repo.

bdittmer:
Ah, that makes sense. Now I know why it barfed so many gems that aren’t mine into my PackageCloud account.

lib/dpl/provider/packagecloud.rb
```diff
@@ -86,7 +86,7 @@ def get_source_files_for(orig_filename)
 
       def push_app
         packages = []
-        glob_args = ["**/*"]
+        glob_args = Array(options.fetch(:package_glob, '**/*'))
```
bwinter: =/

BanzaiMan: @bwinter Could you please elaborate what you mean?

bwinter: Sry, would just like to see this functionality.

BanzaiMan:
This options should be documented.

kylev:
Docs pull request submitted.

PR #295, opened 2015-08-05
Don't auto register PyPI releases for less chance of MITM attacks.

jezdez:
This is a continuation of #253 and a reaction to the comment from the twine author and PyPI admin @dstufft: https://github.com/travis-ci/dpl/pull/288#discussion_r35265268


dstufft:
+1

PR #297, opened 2015-08-06
Improved and updated the Deis provider

aj-may:

PR #300, opened 2015-08-10
Allow the protocol of the deis controller to be specified for the deis provider.

jorihardman:
Deis added https support for the control plane several versions ago, and the current provider does not support an https controller. This adds that support while maintaining backwards compatibility.


jorihardman:
travis-ci/dpl#297 will also fix this, so either way is fine with me. This is currently blocking travis deployment for me.

jorihardman:
Closing this in favor of travis-ci/dpl#297

Issue #94, opened 2014-01-21
Support for Google App Engine (GAE) deployment 


jeanEpek:
We would like to add Google App Engine (GAE) as a supported provider in the deploy section.

We currently use an after_success section but it's not ideal.

Trii:
Using the gcloud tool you can deploy any app regardless of language. It also supports service accounts for auth via `gcloud auth activate-service-account` passing a json key file instead of having to use a user's refresh token.

This is how I implement it:

```
install:
  - if [[ ! -f google-cloud-sdk/path.bash.inc ]]; then
      echo "Downloading google cloud sdk";
      wget https://dl.google.com/dl/cloudsdk/release/google-cloud-sdk.zip;
      unzip -qq google-cloud-sdk.zip;
      google-cloud-sdk/install.sh --usage-reporting false --path-update false --rc-path=~/.bashrc --bash-completion false --override-components=app;
    fi;
  - google-cloud-sdk/bin/gcloud components update preview app --quiet

.... snip

after_success:
- ./travis-ci-deploy.sh
```

Where **travis-ci-deploy.sh* boils down to the following:
```
google-cloud-sdk/bin/gcloud auth activate-service-account \
    --key-file "${GAE_CLIENT_KEY_JSON_FILE}"
google-cloud-sdk/bin/gcloud \
    --project "${GAE_PROJECT_ID}" \
    preview app deploy \
    --version "${GAE_VERSION_LABEL}" \
    --quiet \
    "app.yaml"
```

[gcloud preview app deploy](https://cloud.google.com/sdk/gcloud/reference/preview/app/deploy) reference.

flowlo:
Hi!

I'm using an [`after_success`](https://github.com/coduno/app/blob/master/.travis.yml) similar to @Trii and would love to help improving support for GAE.

Disclaimer: I only use GAE for Go and a bit of Python, also I mainly use managed VMs. The other stuff I do not know very well.

As it seems to me there was a little confusion about Google's tools for GAE. Let me add my portion to this confusion: There is [`appcfg`](https://cloud.google.com/appengine/docs/appcfg) (also `appcfg.py`) which is old (AFAIK) and language agnostic. It is wrapped by other tools (such as `goapp`). I might even suspect that it will fade some day. Different language runtimes have added their flavours to it.

[...]

PR #316, opened 2015-08-22
gae: Reimplement provider for Google App Engine

flowlo:
This now uses Google's Cloud SDK towards a resolution for #94 

I see that tests are missing, but I'd rather like some opinion on the options and install routine beforehand.

And please bear with me, these are the first lines of Ruby in my life.

CC @BanzaiMan @waprin @ajessup @Trii

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -53,10 +53,6 @@ group :gcs do
   gem 'mime-types'
 end
 
-group :gae do
-  gem 'rubyzip'
-end
-
 group :elastic_beanstalk do
   gem 'rubyzip'
   gem 'aws-sdk-v1'
--- a/lib/dpl/provider/gae.rb
+++ b/lib/dpl/provider/gae.rb
@@ -11,6 +11,13 @@ module DPL
       GCLOUD="#{INSTALL}/#{NAME}/bin/gcloud"
 
       def install_deploy_dependencies
+        # FIXME this is a workaround for https://code.google.com/p/google-cloud-sdk/issues/detail?id=228
+        if docker_build == "remote" && !File.exists?("#{Dir.home}/.ssh/google_compute_engine")
+          unless context.shell('ssh-keygen -f ~/.ssh/google_compute_engine -t rsa -N \'\'')
+            error 'Failed to generate SSH key for remote Docker build.'
+          end
+        end
+
         if File.exists? GCLOUD
           return
         end
--- a/lib/dpl/provider/gae.rb
+++ b/lib/dpl/provider/gae.rb
@@ -1,68 +1,83 @@
-require 'digest/sha1'
-require 'open-uri'
-
 module DPL
   class Provider
     class GAE < Provider
       experimental 'Google App Engine'
 
-      # https://developers.google.com/appengine/downloads
-      GAE_VERSION='1.9.17'
-      GAE_ZIP_FILE="google_appengine_#{GAE_VERSION}.zip"
-      SHA1SUM='eec50aaf922d3b21623fda1b90e199c3ffa9e16e'
-      BASE_DIR=Dir.pwd
-      GAE_DIR=File.join(BASE_DIR, 'google_appengine')
-      APPCFG_BIN=File.join(GAE_DIR, 'appcfg.py')
+      BASE='https://dl.google.com/dl/cloudsdk/channels/rapid/'
+      NAME='google-cloud-sdk'
+      EXT='.tar.gz'
+      INSTALL='~'
+      BOOTSTRAP="#{INSTALL}/#{NAME}/bin/bootstrapping/install.py"
+      GCLOUD="#{INSTALL}/#{NAME}/bin/gcloud"
 
-      def self.install_sdk
-        requires 'rubyzip', :load => 'zip'
-        $stderr.puts "Setting up Google App Engine SDK"
+      def install_deploy_dependencies
+        if File.exists? GCLOUD
+          return
+        end
 
-        Dir.chdir(BASE_DIR) do
-          unless File.exists? GAE_ZIP_FILE
-            $stderr.puts "Downloading Google App Engine SDK"
-            File.open(GAE_ZIP_FILE, "wb") do |dest|
-              open("https://storage.googleapis.com/appengine-sdks/featured/#{GAE_ZIP_FILE}", "rb") do |src|
-                dest.write(src.read)
-              end
-            end
-          end
-          sha1sum = Digest::SHA1.hexdigest(File.read(GAE_ZIP_FILE))
-          unless sha1sum == SHA1SUM
-            raise "Checksum did not match for #{GAE_ZIP_FILE}"
-          end
+        $stderr.puts 'Downloading Google Cloud SDK ...'
 
-          unless File.directory? 'google_appengine'
-            $stderr.puts "Extracting Google App Engine SDK archive"
-            Zip::File.open(GAE_ZIP_FILE) do |file|
-              file.each do |entry|
-                entry.extract entry.name
-              end
-            end
-          end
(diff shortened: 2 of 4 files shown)


waprin:
Thanks for doing this, I am no Ruby expert either but I'll try to test it soon.

Gemfile
```diff
@@ -54,7 +54,6 @@ group :gcs do
 end
 
 group :gae do
-  gem 'rubyzip'
 end
```
BanzaiMan: If the non-standard library is no longer required, I think it is better to drop this group altogether.

flowlo: The library was used to unpack the downloaded file prior to this rewrite, so I think it's safe to assume it is not used anymore. Will do so.

Along this note, I added `require`s for `'rubygems/package'` and `'zlib'`. Should they be added here? I don't know anything about dependencies in Ruby.

flowlo:
Justification for dropping SHA1 check:

The zipped file comes with a CRC sum

````
$ gzip -lv google-cloud-sdk.tar.gz
method  crc     date  time           compressed        uncompressed  ratio uncompressed_name
defla e0c0beec Aug 19 17:18             2204347            10680320  79.4% google-cloud-sdk.tar
````

and this sum is honored when decompressing with [`zlib`](http://ruby-doc.org/stdlib-2.2.2/libdoc/zlib/rdoc/Zlib.html)

> The zlib compression library provides in-memory compression and decompression functions, including integrity checks of the uncompressed data.

The file is downloaded via a trusted authenticated channel (HTTP over TLS).

File: Gemfile (hunk 1 above)
Comment (BanzaiMan) [inline Gemfile]: RubyGems and Zlib are standard libraries (http://ruby-doc.org/stdlib-2.1.5/) you can expect to find in functional Ruby runtimes, so there is no need to add them to Gemfile.

Comment (flowlo) [inline Gemfile]: Great, thanks for clarification!

**Comment by BanzaiMan** [comment]
Could you test this with:

```yaml
deploy:
  provider: gae
  edge:
    source: flowlo/dpl
    branch: gae
  ⋮
  on:
    all_branches: true # if necessary
```

flowlo:
Tested it, fails because it [cannot install the Google Cloud SDK](https://travis-ci.org/coduno/api/jobs/77558698#L275), [right here](https://github.com/travis-ci/dpl/pull/316/files#diff-026cdacf16de2151f5cee8b38f51fe36R37).

@BanzaiMan Could you please point me to docs of `context.shell`? I want to get the exit code and standard streams to see what's happening, but do not know how to accomplish that.

lib/dpl/provider/gae.rb
```diff
@@ -1,46 +1,39 @@
...
+          elsif entry.file?
+            File.open target, "wb" do |f|
+              f.print entry.read
             end
+            FileUtils.chmod entry.header.mode, target
           end
         end
+
+        # Bootstrap the Google Cloud SDK.
+        context.shell("CLOUDSDK_CORE_DISABLE_PROMPTS=1 #{Dir.home}/#{NAME}/bin/bootstrapping/install.py --usage-reporting=false --command-completion=false --path-update=false --additional-components=preview")
```
BanzaiMan: Where does this install the `gcloud` executable?

Trii: normally it creates google-cloud-sdk/bin/gcloud from your current directory

BanzaiMan:
The real error is indicated here: https://travis-ci.org/coduno/api/jobs/77558698#L280

Both `#check_auth` and `#push_app` call `GCLOUD`, but only the latter fails when `/home/travis/google-cloud-sdk/bin/gcloud` is invoked.

In between, `#cleanup` is called, and I suspect that *somehow* it is removing `gcloud`. If you add `skip_cleanup: true` under `deploy`, does the deployment work?

lib/dpl/provider/gae.rb
flowlo: It is unzipped to `~`, or `~/google-cloud-sdk` to be precise, with `gcloud` in `~/google-cloud-sdk/bin/gcloud`.

@Trii Yep, this is the default behavior, although I opted to install it to `~` because it is very likely that the current directory is the directory the application being deployed lives in. If the SDK is installed there, it is likely that the whole SDK is copied to App Engine alongside the source code.

lib/dpl/provider/gae.rb
```diff
@@ -1,46 +1,39 @@
...
+        $stderr.puts "Downloading Google Cloud SDK"
 
-          unless File.directory? 'google_appengine'
-            $stderr.puts "Extracting Google App Engine SDK archive"
-            Zip::File.open(GAE_ZIP_FILE) do |file|
-              file.each do |entry|
-                entry.extract entry.name
-              end
+        Gem::Package::TarReader.new(Zlib::GzipReader.open(open(BASE + NAME + EXT, "rb"))).each do |entry|
+          target = File.join(Dir.home, entry.full_name)
```
flowlo: @BanzaiMan This basically defines the install location of the Google Cloud SDK.

flowlo:
I [added `skip_cleanup`](https://github.com/coduno/api/commit/48c78e3381982da28e90747f704a6957ca1b459b) but [still have the same symptom](https://travis-ci.org/coduno/api/jobs/77591609#L278).

BanzaiMan:
@flowlo So bizarre! If you want to inject various debug output (to determine when `/home/travis/google-cloud-sdk/bin/gcloud` disappears), it might be instructive. If you can't do that immediately, I'll take a stab at it some time later.

flowlo:
@BanzaiMan I just threw in some `File.exists`, see [this build](https://travis-ci.org/coduno/api/jobs/77591609) ... And well, TBH I can't really make sense of that. The call to `.../bootstrapping/install.py` does not complain, but still, `gcloud` is missing right after that.

BanzaiMan:
I think I tracked this down to an I/O race condition of sorts.

As currently implemented in the PR, the control is passed on to the next method before the archive is completely read and the content written to disk, or before SDK bootstrapping is complete. This means that, by the time `dpl` gets to run `gcloud`, it is not yet extracted; hence `ENOENT`.

I tried doing it with `system "wget"` & `system "tar xzvf …"`, but this approach, too, has similar problems.

A clean way I've found is to use `wget` and `tar`, but with Open3 and wait for the files we want are written to disk.

BanzaiMan:
See https://github.com/travis-ci/dpl/tree/ha-test-pr-316 for an example build. The deployment fails because the app is not set up for GAE, but you can see that `gcloud` is invoked (https://travis-ci.org/BanzaiMan/travis_production_test/builds/77610416#L193).

flowlo:
I opted for `curl | gzip | tar` over `Gem::Package::TarReader` and `Zlib::GzipReader` for stability. Looks like downloading/unzipping/unarchiving is solid now.

Also, I had to switch to using Docker (`sudo: required` and `services: - docker`), because the newest version of `gcloud` does not respect `--quiet` and would thus hangwaiting for user input. I [posted to their mailing list](https://groups.google.com/forum/#!topic/google-cloud-sdk/DOML82QPp6s) about this, and as soon as it is fixed, remote builds should work automatically.

Provider working smoothly with following `.travis.yml`

````yaml
language: go

sudo: required

go:
- 1.5
- tip

services:
 - docker

notifications:
  email: never

cache:
  directories:
  - "~/google-cloud-sdk/"

before_install:
- $(! $TRAVIS_SECURE_ENV_VARS) || openssl aes-256-cbc -K $encrypted_e505aa14efb0_key
  -iv $encrypted_e505aa14efb0_iv -in secrets.tar.enc -d | tar xv

deploy:
  provider: gae
  edge:
    source: flowlo/dpl
    branch: gae
  project: coduno
  default: false
  docker_build: local
  on:
    all_branches: true
````

PR #326, opened 2015-09-03
Boxfuse provider: Fixed image/version paramater swap

axelfontaine:
This fixes a bug in the previous PR.


axelfontaine:
Note: the build failed for a totally unrelated reason

PR #340, opened 2015-09-24
Fixing typo in Elastic Beanstalk documentation

sebest:


BanzaiMan:
Thank you. Just note that hyphens and underscores are normalized when we process the options.

PR #346, opened 2015-10-08
Fix missing gem "multipart-post" for TestFairy provider

giltsl:


lib/dpl/provider/testfairy.rb
```diff
@@ -70,10 +72,14 @@ def signing_apk(instrumentedFile)
...
       def download_from_url(url)
-        puts "downloading  from #{url} "
+        puts "downloading from #{url} "
+        url = "#{url}?api_key=#{option(:api_key)}"
         uri = URI.parse(url)
         instrumentedFile = Net::HTTP.start(uri.host, uri.port) do |http|
-          resp = http.get(uri.path)
+          resp = http.get "#{uri.path}?#{uri.query}"
+          if resp.code == "302"
+            resp = Net::HTTP.get_response(URI.parse(resp.header['location']))
```
flowlo: [`Net::HTTP` docs on redirection, FYI](http://ruby-doc.org/stdlib-2.2.3/libdoc/net/http/rdoc/Net/HTTP.html#class-Net::HTTP-label-Following+Redirection). You will probably only follow one redirect, which may make sense, or not.

flowlo:
I suggest squashing all those `api_key added *` commits.

lib/dpl/provider/testfairy.rb
gmegidish: Yes. We will only redirect once, it makes complete sense.

gmegidish:
@flowlo squashed! :)

PR #341, opened 2015-10-02
gae: Adapt to gcloud-0.9.80

flowlo:
The new gcloud version deprecates --set-default and introduces --promote (and --no-promote for transition, apparently). This is reflected in the options to the provider. There is no smooth transition from 'default' to 'no_promote' as the provider is experimental.

Also, the release notes hint that --no-stop-previous-version will become available as soon as --promote kicks in. This is also exposed as an option.

See https://dl.google.com/dl/cloudsdk/release/RELEASE_NOTES

PR #350, opened 2015-10-16
gae: Adapt to gcloud-0.9.82

flowlo:
Supersedes #341


step76:
I'm sorry to say that this doesn't work: the plugin is still trying to install the "preview" component, but since 0.9.80 the "preview" component has disappeared and now should be installed the "app" component.

PR #356, opened 2015-10-20
Fix undefined local variable errors on deploy

clayreimann:
As see in build: https://travis-ci.org/yahoo/elide/builds/86486285

### DIFF
--- a/lib/dpl/provider/bintray.rb
+++ b/lib/dpl/provider/bintray.rb
@@ -149,6 +149,7 @@ module DPL
         if code == 201 || code == 200
           return true
         end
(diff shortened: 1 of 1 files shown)


clayreimann:
Hey @BanzaiMan how long till this is useable? Trying to diagnose why my deploy fails.

BanzaiMan:
You can try it with `edge: true` until there is a release.

clayreimann:
Cool. What's your release cadence?

BanzaiMan:
We don't have a set schedule.

BanzaiMan:
I've pushed 1.8.7.

PR #357, opened 2015-10-30
Implement slot support for azure web apps deployment

dzonatan:
Azure uses different git url format for slot deployment (implemented in this PR):
```
https://{SLOT}.scm.azurewebsites.net:443/{SITE}.git
```

If **slot** config is not passed then the git url stays the same as it was previously:
```
https://{SITE}.scm.azurewebsites.net:443/{SITE}.git
```

This is not a breaking-change!

More about staging development in azure [read here](https://azure.microsoft.com/en-us/documentation/articles/web-sites-staged-publishing/).
Sorry but I don't have any build links to show since I'm using this provider only in private project...

### DIFF
--- a/README.md
+++ b/README.md
@@ -444,12 +444,13 @@ You first need to create an [Atlas account](https://atlas.hashicorp.com/account/
 #### Environment variables:
 
  * **AZURE_WA_SITE** Web App Name. Used if the `site` option is omitted.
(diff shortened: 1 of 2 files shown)


BanzaiMan:
Sorry for the inconvenience. Due to the broken specs on master (which has now been fixed), your PR failed. Could you rebase this PR? Thank you!

dzonatan:
Done that.

PR #361, opened 2015-11-16
Azure Deployment: skip_cleanup

### LINKED ISSUE #360: Having difficulty with Azure deployment
[Here is](https://github.com/Enttoi/enttoi-ui-web/blob/master/.travis.yml) my yml file. I spent quit a lot time on it - it seems that the parameter skip_cleanup in deploy section is ignored, as result all artifacts produced during build are not deployed. 

I tried deploying via script by pushing to remote and it is working (/dist presented in source and destination).

felixrieseberg:
Closes #360

### DIFF
--- a/lib/dpl/provider/azure_webapps.rb
+++ b/lib/dpl/provider/azure_webapps.rb
@@ -30,6 +30,13 @@ module DPL
       def push_app
         log "Deploying to Azure Web App '#{config['slot'] || config['site']}'"
 
+        if !!options[:skip_cleanup]
+          log "Skipping Cleanup"
+          context.shell "git checkout master"
+          context.shell "git add . --all --force"
+          context.shell "git commit -m \"Skip Cleanup Commit\""
+        end
+
         if !!options[:verbose]
           context.shell "git push --force --quiet #{git_target} master"
         else


lib/dpl/provider/azure_webapps.rb
```diff
@@ -29,6 +29,9 @@ def check_auth
       def push_app
         log "Deploying to Azure Web App '#{config['site']}'"
 
+        if options[:skip_cleanup]
+          context.shell "git add --all --force > /dev/null 2>&1"
+
```
BanzaiMan: I don't think this is valid syntax.

felixrieseberg:
Updated.

lib/dpl/provider/azure_webapps.rb
```diff
@@ -29,6 +29,10 @@ def check_auth
       def push_app
         log "Deploying to Azure Web App '#{config['site']}'"
 
+        if options[:skip_cleanup]
+          context.shell "git add --all --force > /dev/null 2>&1"
+          context.shell "git commit -m \"Skip Cleanup Commit\" > /dev/null 2>&1"
+
```
BanzaiMan: Hmmm. It's still wrong. There should be `end` here.

felixrieseberg:
Ugh, that's what I get for trying to quickly do this from a phone. It's now updated and actually tested (https://travis-ci.org/felixrieseberg/dpl-azure-test).

jenyayel:
Works as expected now. Thank you @BanzaiMan and @felixrieseberg.

PR #362, opened 2015-11-17
Fixes #207 Modulus deploys failing without error

### LINKED ISSUE #207: Modulus deploys failing without error
When attempting to deploy to Modulus, the deploy step is empty and exits with 0 although my project isn't deployed. Is there a way to see if there are any error from the deploy provider module or any indication as to why the deploy never happens?

jakehewitt:
The command was failing with an exit status of 127. The problem was
trying to set the environment variable through context.shell, this
should be done using Ruby's ENV variable

@BanzaiMan


lib/dpl/provider/modulus.rb
```diff
@@ -16,7 +16,8 @@ def needs_key?
       end
 
       def push_app
-        context.shell "MODULUS_TOKEN=#{option(:api_key)} modulus deploy -p #{option(:project_name)}"
+        ENV['MODULUS_TOKEN']="#{option(:api_key)}"
```
BanzaiMan: Does `ENV` affect how commands are executed in `context.shell`? I tend to think it does not. Has this been tested?

Perhaps we can do with `context.shell "env MODULUS_TOKEN=#{option(:api_key)} modulus …"`.

jakehewitt: @BanzaiMan Yeah that works too and its cleaner. I redid the PR as #363 and closed this one

PR #364, opened 2015-11-19
Changed the additional components to be installed

step76:
The right component to be installed is "app" and not "preview", the "preview" component has disappeared.

This should fix #350


flowlo:
Thank you for spotting and fixing this :bow: LGTM.

@BanzaiMan The provider currently is broken (my fault), so it would be great if we could get this merged in a timely manner. Thanks!

PR #365, opened 2015-11-21
Pin mime-types to 2.x for mri 1.9

fasterthanlime:
Upgrading to mri 2.x would be preferrable, but in the meantime, this should help stop the bleeding on https://github.com/travis-ci/travis-ci/issues/5145

### DIFF
--- a/lib/dpl/provider/gcs.rb
+++ b/lib/dpl/provider/gcs.rb
@@ -4,7 +4,7 @@ module DPL
   class Provider
     class GCS < Provider
       requires 'gstore'
-      requires 'mime-types'
+      requires 'mime-types', version: '~> 2.0'
 
       def needs_key?
         false
--- a/lib/dpl/provider/s3.rb
+++ b/lib/dpl/provider/s3.rb
@@ -4,7 +4,7 @@ module DPL
   class Provider
     class S3 < Provider
       requires 'aws-sdk', version: '>= 2.0.22'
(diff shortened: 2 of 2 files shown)


BanzaiMan:
Looks like `ohai` requires `gem 'mime-types', '~> 1.16'`. (https://travis-ci.org/travis-ci/dpl/jobs/92456598#L129).

fasterthanlime:
hmm that second commit might be a bit thoughtless :( why would mime-types-data 3.x even be installed in the build I posted, if ohai requires mime-types 1.16?

I'll run some tests locally before messing with dependencies some more

fasterthanlime:
@BanzaiMan bundler stopped complaining after `0a641aff42b0f2c4a627f74a9b6ba89489bcbf66` but still seeing 4 failures in the logs — any ideas where those come from? I can't seem to figure it out :(

fasterthanlime:
Alright, force-pushed another fix to my PR branch, finally realized dpl has another set of dependencies it installs itself at runtime, depending on the provider.

This makes my deploys work ([proof](https://travis-ci.org/itchio/elevate/jobs/92468385)) — but might very well break dpl on ruby 2.x and above. How can I check that? It seems that on travis, `dpl` is run with 1.9.3 no matter what you `rvm switch` to.

BanzaiMan:
Thank you!

PR #367, opened 2015-11-22
lock down to 2.6.2 to keep rvm 1.9 compatible

zhengshaolin:
mime-types upgrade to 3.0.0 on Nov 21st, which cause the deploy to s3 failed because 3.0.0 is not compatible with rvm 1.9. So we lock down to an earlier version.


BanzaiMan:
Thanks for the PR. I am closing this in favor of #365.

PR #379, opened 2015-12-16
Fixed provider anynines to work in TravisCI

benjaminguttmann-avtq:
Now cf api endpoint is set correctly, furthermore user can login and set organization and spaces.

### DIFF
--- a/lib/dpl/provider/anynines.rb
+++ b/lib/dpl/provider/anynines.rb
@@ -4,8 +4,8 @@ module DPL
 
       def check_auth
         initial_go_tools_install
-        context.shell "cf api https://api.de.a9s.eu"
-        context.shell "cf login --u #{option(:username)} --p #{option(:password)} --o #{option(:organization)} --s #{option(:space)}"
+        context.shell "./cf api https://api.de.a9s.eu"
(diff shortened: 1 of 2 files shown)


BanzaiMan:
Sorry for the delay. What is the problem this PR aims to solve? Do you have a build log URL that shows this problem? Has this PR been tested?

benjaminguttmann-avtq:
If you want to deploy something to the provider anynines you get the following log:

cf
dpl.3
Deploying application
No API endpoint set. Use 'cf login' or 'cf api' to target an endpoint.
FAILED
No org and space targeted, use 'cf target -o ORG -s SPACE' to target an org and space
Logging out...
OK

With the changes the provider should work correctly.

BanzaiMan:
@benjaminguttmann-avtq Where would I see this error message?

Perhaps I am missing something, but I fail to see how this PR would resolve any error that results in a message like that. This PR changes the location of the `cf` executable to the current directory, and current directory only.

A likely cause of your error is missing arguments, perhaps because you tried encrypting values but that failed for one reason or another.

Please follow instructions in https://github.com/travis-ci/dpl/blob/master/TESTING.md to test this PR, and please link to the build that resolves deployment issue.

benjaminguttmann-avtq:
So, here is the link to the build with my changes: https://travis-ci.org/benjaminguttmann-avtq/ruby-sinatra-example-app/builds/101559287

PR #380, opened 2015-12-18
Add debug options (verbose and retry) for Heroku upload

BanzaiMan:
- `verbose: true` passes `-vvv` to `curl`
- `retry: N` passes `--retry N` to `curl`

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -63,6 +63,7 @@ end
 
 group :bitballoon do
   gem 'bitballoon'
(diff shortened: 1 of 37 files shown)


BanzaiMan:
`curl -vvv` exposes most HTTP exchanges, including AWS requests with the access key id and the signature. This ***could*** be problematic, so it is probably best to filter output.

BanzaiMan:
Stale.

PR #386, opened 2016-01-07
Allow Github release provider to overwrite files

lazyatom:
Sometimes we want to updating an existing release with a new file that
has the same name. While best practice for user-facing releases is to
make a brand new release, there are some use cases where releases are
used as well known endpoints to grab the latest code.

This commit adds support for an `overwrite` option for the releases
provider, which will upload files even if others with the same name
are present on the release.

### DIFF
--- a/lib/dpl/provider/releases.rb
+++ b/lib/dpl/provider/releases.rb
@@ -100,22 +100,23 @@ module DPL
         end
 
         files.each do |file|
-          already_exists = false
+          existing_url = nil
           filename = Pathname.new(file).basename.to_s
           api.release(release_url).rels[:assets].get.data.each do |existing_file|
             if existing_file.name == filename
-              already_exists = true
+              existing_url = existing_file.url
             end
           end
-          if already_exists
-            log "#{filename} already exists, skipping."
-          else
+          if !existing_url || (existing_url && options[:overwrite])
+            api.delete_release_asset(existing_url) if existing_url
             content_type = MIME::Types.type_for(file).first.to_s
             if content_type.empty?
               # Specify the default content type, as it is required by GitHub
               content_type = "application/octet-stream"
             end
             api.upload_asset(release_url, file, {:name => filename, :content_type => content_type})
+          else
+            log "#{filename} already exists, skipping."
           end
         end
 
--- a/lib/dpl/provider/releases.rb
+++ b/lib/dpl/provider/releases.rb
@@ -108,7 +108,6 @@ module DPL
             end
           end
           if !existing_url
-            log "#{filename} doesn't exist, uploading."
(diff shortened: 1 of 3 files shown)


lazyatom:
Actually, hold off with this; it doesn't actually work, because Github itself complains about the existing asset. I'm working on a version that explicitly removes the asset now.

lazyatom:
I'm having some trouble with Octokit's `delete_release_asset` method. I've filed an issue at octokit/octokit.rb#709

lazyatom:
OK! I think this is ready for consideration now. You can see an example build running here: https://travis-ci.org/lrug/lrug.org/builds/100860504

lib/dpl/provider/releases.rb
```diff
@@ -100,22 +100,23 @@ def push_app
...
             if existing_file.name == filename
-              already_exists = true
+              existing_url = existing_file.url
             end
           end
-          if already_exists
-            log "#{filename} already exists, skipping."
-          else
+          if !existing_url || (existing_url && options[:overwrite])
+            api.delete_release_asset(existing_url) if existing_url
```
BanzaiMan: It might be a good idea to log that existing files are overwritten.

lazyatom: Added in 4464031; it required a little more refactoring to achieve without duplication.

BanzaiMan:
Looks like my work on Releases introduced a conflict. Please resolve the conflict, and I'll take another look. Thanks.

lazyatom:
Done (3a1bbea)

lib/dpl/provider/releases.rb
```diff
@@ -100,27 +100,36 @@ def push_app
...
           api.release(release_url).rels[:assets].get.data.each do |existing_file|
             if existing_file.name == filename
-              already_exists = true
+              existing_url = existing_file.url
             end
           end
-          if already_exists
-            log "#{filename} already exists, skipping."
+          if !existing_url
+            log "#{filename} doesn't exist, uploading."
```
BanzaiMan: Is this desirable? It seems to me that this is the default behavior, and this would be mostly noise.

lazyatom: I don't feel strongly about it either way - happy to remove it if you'd prefer.

BanzaiMan: Sorry about the delay. Yes, please remove this.

lazyatom: Done!

BanzaiMan:
Looks like the spec needs to be dropped.

lazyatom:
Ah, apologies for that. Should be done now, and I've rebased against your latest master.

PR #394, opened 2016-01-20
ElasticBeanstalk option for only_create_app_version

georgantasp:
Hey Gents,

These commit add an only_create_app_version option to the elastic beanstalk provider. This makes it so the version is uploaded without making it active.

Thanks.

### DIFF
--- a/lib/dpl/provider/elastic_beanstalk.rb
+++ b/lib/dpl/provider/elastic_beanstalk.rb
@@ -27,6 +27,10 @@ module DPL
       def check_app
       end
 
+      def only_create_app_version
+        options[:only_create_app_version]
+      end
+
       def push_app
         create_bucket unless bucket_exists?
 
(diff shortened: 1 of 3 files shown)


lib/dpl/provider/elastic_beanstalk.rb
```diff
@@ -27,6 +27,10 @@ def check_auth
       def check_app
       end
 
+      def only_create_app_version
+        options[:only_create_app_version] || false
```
BanzaiMan: `|| false` seems unnecessary, as the value defaults to `nil` is the key doesn't exist.

BanzaiMan:
Could you also squash these commits, and add a document on README, please?

georgantasp:
Squashed (i think?) and confirming the README was updated as well.

Thanks.

georgantasp:
Is there anything else I can do to have this merged?

PR #396, opened 2016-01-31
Added ability to specifc PyPI config options as environment variables.

jezdez:
This would allow setting the PyPI config variables conditionally in a pre step for example.


jezdez:
woohoo! thanks :cake: :heart_eyes: :fireworks:

Issue #352, opened 2015-10-16
Public key permission denied on deploy to Deis


daanemanz:
When deploying an application to Deis, the Deis platform rejects the key that's generated by `dpl` during the deploy phase.

Running Deis 1.11.1, please see the deploy log below:

```
Installing deploy dependencies
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100 1800k  100 1800k    0     0  80349      0  0:00:22  0:00:22 --:--:-- 69365
Verifying archive integrity... All good.
Uncompressing Deis CLI  100%  
deis is in the current directory. Please
move deis to a directory in your search PATH.
See http://docs.deis.io/ for documentation.

Preparing deploy
Logged in as ci
=== example-api Application
updated:  2015-10-16T15:29:32UTC
uuid:     3713968d-2fa7-4372-99ed-c64301150683
created:  2015-10-16T14:53:08UTC
url:      example-api.example.com
owner:    ci
id:       example-api
=== example-api Processes
--- web:
web.1 up (v4)
=== example-api Domains
Generating public/private rsa key pair.
Your identification has been saved in .dpl/id_rsa.
Your public key has been saved in .dpl/id_rsa.pub.
The key fingerprint is:
a8:41:46:ee:82:67:8d:50:76:eb:d9:44:1a:bb:91:bb testing-worker-linux-docker-9a70e625-3370-linux-1
The key's randomart image is:
+--[ RSA 2048]----+
|  o + .          |
| o + B           |
|.   X .          |
| o B B .         |
|. = O o S        |
| o . +           |
|    E            |
|                 |
|                 |
+-----------------+
Uploading id_rsa.pub to deis... done
Git remote deis added
Saved working directory and index state WIP on (no branch): <commit>
HEAD is now at <commit>

Deploying application
Warning: Permanently added '[deis.example.com]:2222' (ECDSA) to the list of known hosts.
Permission denied (publickey).
fatal: Could not read from remote repository.
Please make sure you have the correct access rights
and the repository exists.
Removing testing-worker-linux-docker-9a70e625-3370-linux-1 SSH Key... done
Already up-to-date!
# HEAD detached at <commit>
# Untracked files:
#   (use "git add <file>..." to include in what will be committed)
#
#   .dpl/
#   deis
#
nothing added to commit but untracked files present (use "git add" to track)
Dropped refs/stash@{0} (<commit>)
[...]

PR #402, opened 2016-02-10
Add a check to wait for the SSH key to start working

mattk42:
This is for #352. Although for some reason just using master seemed to work, it does seem that there is a timing issue that should be explicitly solved.

The problem seems to be that Deis reports ssh keys are added slightly before they can actually be used for a code push. In my testing it seems to take < 2 seconds for the key to be usable.

This code just gets the deis remote information from the git config file (for some reason `git remove -v show deis` was not working) and attempts to SSH into the remote until it gets a 'PTY allocation request failed message'. That doesn't seem like the cleanest way of going about it, but since you can't actually SSH in it was the only test I could come up with. Seems to work for me.

Edit: I realize that this doesn't pass tests right now, if this implementation looks ok then I will go and fix the tests.

### DIFF
--- a/lib/dpl/provider/deis.rb
+++ b/lib/dpl/provider/deis.rb
@@ -44,6 +44,31 @@ module DPL
         unless context.shell "./deis git:remote --app=#{option(:app)}"
           error 'Adding git remote failed.'
         end
+
+        wait_for_git_access
+      end
+
+      def wait_for_git_access()
+        retry_count=0
+        max_retries=30
+
+        #Parse out the git remote host and port
+        git_remote=`cat .git/config  | grep -A1 '[remote] "deis"' | tail -n 1 | sed 's/.*url = //'`
+        remote_uri=git_remote.split("ssh://")[1].split("/")[0]
+        remote_host=remote_uri.split(":")[0]
+        remote_port=remote_uri.split(":")[1]
+        puts "Git remote is #{remote_host} at port #{remote_port}"
+
+        #Try and connect to the github remote via ssh.
+        while retry_count < max_retries
+          puts "Waiting for ssh key to propagate..."
+          if context.shell "#{context.env['GIT_SSH']} #{remote_host} -p #{remote_port}  2>&1 | grep -c 'PTY allocation request failed' > /dev/null"
+            puts "SSH connection established."
+            break
+          end
+          retry_count += 1
+          sleep(1)
+        end
       end
 
       def remove_key
--- a/lib/dpl/provider/deis.rb
+++ b/lib/dpl/provider/deis.rb
@@ -1,6 +1,9 @@
 module DPL
   class Provider
     class Deis < Provider
+      
+      requires 'git'
+
       def install_deploy_dependencies
         context.shell "curl -sSL http://deis.io/deis-cli/install.sh | sh -s #{option(:cli_version)}"
       end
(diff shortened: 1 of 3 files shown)


lib/dpl/provider/deis.rb
```diff
@@ -44,6 +44,31 @@ def setup_git_ssh(path, key_path)
...
+
+        wait_for_git_access
+      end
+
+      def wait_for_git_access()
+        retry_count=0
+        max_retries=30
+
+        #Parse out the git remote host and port
+        git_remote=`cat .git/config  | grep -A1 '[remote] "deis"' | tail -n 1 | sed 's/.*url = //'`
```
BanzaiMan: It's bordering on style, but there is no reason to pipe `cat` into `grep`; `grep -A1 '[remote] "deis"' .git/config` should be sufficient.

lib/dpl/provider/deis.rb
```diff
@@ -44,6 +44,31 @@ def setup_git_ssh(path, key_path)
...
+        remote_uri=git_remote.split("ssh://")[1].split("/")[0]
```
BanzaiMan: This assumes that `git_remote` is nonempty; which might be true in practice (that there is a remote named `deis`), but as you can see in the failing specs, it may not be true.

lib/dpl/provider/deis.rb
```diff
@@ -44,6 +44,31 @@ def setup_git_ssh(path, key_path)
...
+        remote_host=remote_uri.split(":")[0]
+        remote_port=remote_uri.split(":")[1]
```
BanzaiMan: This is more succinctly written as:

```ruby
remote_host, remote_port = remote_uri.split(":")
```

lib/dpl/provider/deis.rb
mattk42: I agree, I came back to look at this today and noticed that this is kind of hacky anyway. I just started a branch locally that uses https://github.com/schacon/ruby-git so that it can be mocked out in the specs. If the git remote is empty in any situation outside of rspec I think that there would be something wrong already.

spec/provider/deis_spec.rb
```diff
@@ -67,6 +67,15 @@
       expect(provider.context).to receive(:shell).with(
         './deis git:remote --app=example'
       ).and_return(true)
+
+      expect(provider.context).to receive(:shell).with(
+        /grep -c 'PTY allocation request failed'/
```
BanzaiMan: `:shell` is invoked with a lot longer argument.

lib/dpl/provider/deis.rb
mattk42: I fixed the test to mock out these calls so that it gets a host/port. I think in any real-world scenario the `deis git:remote --app` call that happens before this should set up the deis remote. If it doesn't then the deploy is going to bomb out somewhere either way.

lib/dpl/provider/deis.rb
mattk42: Thanks, I went ahead and moved to that.

spec/provider/deis_spec.rb
mattk42: It is, but since most of the rest of it was dynamic and this part was pretty unique I wasn't concerned with looking for the whole line. I can come up with some regex to match it all if that is still preferred though.

PR #414, opened 2016-03-08
Remove cloudControl and dependent providers due to bancruptcy

stefan-kolb:
All services and whitelabel PaaS provided by cloudControl are offline.
This includes cloudControl, exoscale, dotCloud.
https://twitter.com/cloudcontrolled/status/699530071481196544


stefan-kolb:
@BanzaiMan Do you consider merging this PR or aren't you interested?

sfriesel:
@BanzaiMan the exoscale PaaS was *not* shut down (https://community.exoscale.ch/documentation/apps/) and our customers now lost access to this travis integration. Would you consider reverting this merge? Or should I rather re-introduce the exoscale class without the cloudcontrol base? It's still the same underlying API.

stefan-kolb:
Interesting information! the main site of Exoscale doesn't even mention Exoscale Apps anymore: https://www.exoscale.ch/. Maybe this will be shutdown in the near future? Anyhow, maybe we need to revert parts of the commit for now then.

PR #410, opened 2016-02-27
Add cloud-build option to GAE 

jonathan-s:
When deploying it'll allow to let the container builder to perform the docker builds as per google docs. https://cloud.google.com/sdk/gcloud/reference/config/set

use_cloud_build
    If True, use the Container Builder API to perform docker builds, rather than a temporary VM. See https://cloud.google.com/container-builder/docs/ for more information.

PR #419, opened 2016-03-24
gae: Use Container Builder by default, add docs

flowlo:
Reading a [recent announcement from the Google Cloud Platform team](https://groups.google.com/forum/#!topic/google-cloud-sdk/DuOdQPy9PoQ), I understand that Cloud Container Builder should be the default way to build Docker images as part of the deployment process, and developers will see that change automatically without altering any configs.

So the deployment provider should reflect this as well, e.g. promote Cloud Container Builder to opt-out.

Building on #410 I also added a line of documentation that introduces the parameter with it's new semantics.


flowlo:
Succeeded by #463

PR #420, opened 2016-03-24
Launchpad support

Nef10:
This PR adds support to trigger a code import on [launchpad.net](http://launchpad.net). This is useful to, for example, build and host Ubuntu packages.

Example [config](https://github.com/HPI-BP2015H/test/blob/master/.travis.yml#L3-L14) and [build](https://travis-ci.org/HPI-BP2015H/test/builds/118020367).

/cc: @timfel

### DIFF
--- a/lib/dpl/provider/launchpad.rb
+++ b/lib/dpl/provider/launchpad.rb
@@ -1,0 +1,33 @@
+require 'net/http'
+require 'net/https'
+
+module DPL
+  class Provider
+    class Launchpad < Provider
+
+      def check_auth
+      end
+
+      def needs_key?
+        false
+      end
+
+      def push_app
+        http = Net::HTTP.new('api.launchpad.net', 443)
+        http.use_ssl = true
+        req = Net::HTTP::Post.new("/1.0/" + options[:slug] + "/+code-import")
+        req.set_form_data('ws.op' => 'requestImport')
+        req['Authorization'] =
+          'OAuth oauth_consumer_key="Travis%20Deploy", ' +
+          'oauth_nonce="' + rand(36**32).to_s(36) + '",' +
+          'oauth_signature="%26' + options[:oauth_token_secret] + '",' +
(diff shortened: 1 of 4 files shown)


jchromik:
@BanzaiMan could you have a look?

lib/dpl/provider/launchpad.rb
```diff
@@ -0,0 +1,48 @@
...
+      end
+
+      def needs_key?
+        false
+      end
+
+      def push_app
+        response = api_call('/1.0/' + options[:slug] + '/+code-import', {'ws.op' => 'requestImport'})
+        error('Deploy failed! Launchpad credentials invalid. ' + response.code.to_s) if response.code == '401'
+        error('Error: ' + response.code + ' ' + response.body) if response.code != '200'
```
BanzaiMan: Is '200' the only possible success status? Do we follow redirects? How about other 2XX statuses?

Perhaps we want to look at `Net::HTTPSuccess` instead.

lib/dpl/provider/launchpad.rb
```diff
@@ -0,0 +1,48 @@
...
+      private
+
+        def api_call(path, data)
+          req = Net::HTTP::Post.new(path)
+          req.set_form_data(data)
+          req['Authorization'] = get_authorization_header
+          return @http.request(req)
+        end
+
+        def get_authorization_header
```
BanzaiMan: `get_` prefix isn't needed here.

BanzaiMan: And to be more nit picking, the value this method returns is not a header, either. Perhaps just `authorization`.

Nef10:
@BanzaiMan adjusted. We do not need to follow redirects because the repo url on launchpad cannot be changed, so there is no redirect.

Nef10:
thanks for merging, it [works](https://travis-ci.org/HPI-BP2015H/test/builds/122446614#L148) :) 
@timfel make sure to include `edge: true` in your deploy section

PR #426, opened 2016-03-31
Let Travis deploy to Surge.sh

mathiasrw:
Lets support deploying to www.surge.sh - a static web publishing service. 

testrun: https://travis-ci.org/mathiasrw/travis-loves-surge/builds/119932444

---
This change is [Reviewable](https://reviewable.io/reviews/travis-ci/dpl/426)

### DIFF
--- a/lib/dpl/provider/surge.rb
+++ b/lib/dpl/provider/surge.rb
@@ -2,10 +2,9 @@ module DPL
   class Provider
     class Surge < Provider
       npm_g 'surge'
-      context.shell "git rev-parse && cd \"$(git rev-parse --show-toplevel)\"" # Go to repo root 
 
       def project
-        File.expand_path("./" + (options[:project] || '') )
+        File.expand_path(context.env['TRAVIS_BUILD_DIR'] + "/" + (options[:project] || '') )
       end
 
       def domain
--- a/lib/dpl/provider/surge.rb
+++ b/lib/dpl/provider/surge.rb
@@ -2,6 +2,7 @@ module DPL
   class Provider
     class Surge < Provider
       npm_g 'surge'
+      context.shell "git rev-parse && cd \"$(git rev-parse --show-toplevel)\"" # Go to repo root 
 
       def project
         File.expand_path("./" + (options[:project] || '') )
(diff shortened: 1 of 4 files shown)


lib/dpl/provider/surge.rb
```diff
@@ -0,0 +1,33 @@
...
+      def project
+        File.expand_path("./" + (options[:project] || '') )
+      end
+
+      def domain
+        options[:domain] || ''
+      end
+
+      def check_auth
+		raise Error, "Please add SURGE_TOKEN in Travis settings (get your token with 'surge token')" unless context.env['SURGE_TOKEN']
```
BanzaiMan: Please use consistent indentation here.

lib/dpl/provider/surge.rb
```diff
@@ -0,0 +1,33 @@
+module DPL
+  class Provider
+    class Surge < Provider
+      npm_g 'surge'
+
+      def project
+        File.expand_path("./" + (options[:project] || '') )
```
BanzaiMan: Does this work if the working directory changes during the build? (I'm not familiar with `surge`, so this is a genuine question.)

lib/dpl/provider/surge.rb
```diff
@@ -0,0 +1,33 @@
...
+        options[:domain] || ''
+      end
+
+      def check_auth
+		raise Error, "Please add SURGE_TOKEN in Travis settings (get your token with 'surge token')" unless context.env['SURGE_TOKEN']
+        raise Error, "Please add SURGE_LOGIN in Travis settings (its your email)" unless context.env['SURGE_LOGIN']
+      end
+
+      def check_app
+      	raise Error, "Please set a valid project folder path in .travis.yml under deploy: project: myPath" unless File.directory?(project)
```
BanzaiMan: I'd avoid using the postfix conditionals here, since it is very had to see them with such a long message.

lib/dpl/provider/surge.rb
```diff
@@ -0,0 +1,33 @@
...
+      	raise Error, "Please set domain as .travis.yml under deploy: project: myDomain (or in a CNAME file in the repo project folder)" unless ''!=domain || File.exist?("#{project}/CNAME")
```
BanzaiMan: Ditto.

Personally I prefer `domain.empty?` to `'' != domain`.

Also, `unless` with compound conditions are hard to understand. I suggest

```ruby
if domain.empty? && ! File.exist?("#{project}/CNAME")
```

mathiasrw:
@BanzaiMan thank you for very specifik inputs on how to make the code better. Its awesome you  took the time to state how to make it better.

Ill be back !

lib/dpl/provider/surge.rb
mathiasrw: No - if working directory changes during the build this will not work. 

I had the impression that working directory would reset to repo root when deploying. This might not be the case. 

Are you aware of how to identify repo root?

mathiasrw:
Link to successful deploy: https://travis-ci.org/mathiasrw/travis-loves-surge/builds/121234586

lib/dpl/provider/surge.rb
```diff
@@ -0,0 +1,34 @@
+module DPL
+  class Provider
+    class Surge < Provider
+      npm_g 'surge'
+      context.shell "git rev-parse && cd \"$(git rev-parse --show-toplevel)\"" # Go to repo root
```
BanzaiMan: Avoid changing working directories without user's knowledge. There are other tasks that could run after the deployment, and they may assume to be in a certain directory.

Use `pushd` and `popd`.

BanzaiMan: Furthermore, I am not sure if this is the right place to change directories. This part is executed when the file is loaded, which doesn't seem a good idea.

PR #431, opened 2016-04-13
Allow for runtime to be an option for lambdas

### LINKED ISSUE #397: Lambda provider doesn't support Java runtime
Following instructions here:
https://github.com/travis-ci/dpl#lambda

This is my .travis.yml:
https://github.com/Sage-Bionetworks/Bridge-EX-Scheduler/blob/a67afea3198df50ccb62a95761204588e05a3e6b/.travis.yml

Looking at the provider, it looks like the runtime is hardcoded to always be Node.js:
https://github.com/travis-ci/dpl/blob/master/lib/dpl/provider/lambda.rb

icco:
This should fix #397. Amazon Lambdas now support four different runtimes.


icco:
cc @dtothefp

PR #432, opened 2016-04-14
ElasticBeanstalk provider: finish the build after EB env update completes

lushchick:
This PR adds a possibility to finish the build after ElasticBeanstalk environment update completes, if `wait_until_deployed` config option provided. It also prints out EB events messages during the build.

Tested both failing and successful builds:
Failing build (reason - broken .ebextensions config file): [23984152](https://travis-ci.com/planorama/PlanoLive/builds/23984152)
Successful build: [37741814](https://travis-ci.com/planorama/PlanoLive/jobs/37741814)

### DIFF
--- a/lib/dpl/provider/elastic_beanstalk.rb
+++ b/lib/dpl/provider/elastic_beanstalk.rb
@@ -1,3 +1,5 @@
+require 'time'
+
 module DPL
   class Provider
     class ElasticBeanstalk < Provider
@@ -32,6 +34,7 @@ module DPL
       end
 
       def push_app
+        @start_time = Time.now
         create_bucket unless bucket_exists?
 
         if options[:zip_file]
@@ -45,6 +48,7 @@ module DPL
         version = create_app_version(s3_object)
         if !only_create_app_version
           update_app(version)
+          wait_until_deployed if options[:wait_until_deployed]
         end
       end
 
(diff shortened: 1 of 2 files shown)


lib/dpl/provider/elastic_beanstalk.rb
```diff
@@ -138,6 +142,40 @@ def create_app_version(s3_object)
         eb.create_application_version(options)
       end
 
+      # Wait until EB environment update finishes
+      def wait_until_deployed
+        success = true
```
BanzaiMan: Somewhat pedantic, but I think we should assume `success = false`, then set it to `true` when the upload did succeed.

andreylushchick-planorama: I had to change the logic a bit, because EB may send events like `info, error, error, info`, and then success flag would be set to `true, false, false, true`, so would be true despite the errors. So i've added errors counter instead. If there's 1 or more error events - build fails.

lushchick:
@BanzaiMan any chance to merge this PR please? :)
I'm using edge build option with my repo/branch, but for some projects i have 2 EB environments to be deployed, thus 2 providers. And i can't use edge for both of them, because for the 2nd provider travis clones edge DPL branch into the same directory and build fails with git error `fatal: destination path already exists and is not an empty directory.` So i can't wait to remote "edge" option for my builds :)

BanzaiMan:
Could you try testing again?

Thank you.

lushchick:
It works well now, thanks! 🎉

BanzaiMan:
Ah, sorry. I forgot to ask you for specs. Add them, and I'll merge.

andreylushchick-planorama:
Ok, added the specs.

davesag:
Looks like exactly what I need :-)

andreylushchick-planorama:
@BanzaiMan can you please have a look? :)

BanzaiMan:
Please test again with `edge: true`.

andreylushchick-planorama:
Tested, everything went well: [26655506](https://travis-ci.com/planorama/PlanoExport/builds/26655506)

PR #438, opened 2016-04-27
Azure web apps: Enabling push to an empty repository

A92hm:
The script below was unable to push the build to a freshly created Azure Web app, because 
an empty Azure website doesn’t have a Git `HEAD` yet so the deployment fails.@felixrieseberg

Here is a link to a successful build and deploy after my changes.
https://travis-ci.org/A92hm/travis-azure

### DIFF
--- a/lib/dpl/provider/azure_webapps.rb
+++ b/lib/dpl/provider/azure_webapps.rb
@@ -38,9 +38,9 @@ module DPL
         end
 
         if !!options[:verbose]
-          context.shell "git push --force --quiet #{git_target} HEAD:master"
+          context.shell "git push --force --quiet #{git_target} HEAD:refs/heads/master"
         else
-          context.shell "git push --force --quiet #{git_target} HEAD:master > /dev/null 2>&1"
+          context.shell "git push --force --quiet #{git_target} HEAD:refs/heads/master > /dev/null 2>&1"
         end
       end
     end
--- a/spec/provider/azure_webapps_spec.rb
+++ b/spec/provider/azure_webapps_spec.rb
@@ -1,0 +1,95 @@
+require 'spec_helper'
+require 'dpl/provider/anynines'
+
+
+describe DPL::Provider::AzureWebApps do
+  subject :provider do
+    described_class.new(DummyContext.new,
+                        :username => 'myUsername',
+                        :password => 'myPassword',
+                        :site => 'myWebsite',
+                        :slot => 'mySlot')
+  end
+
+  subject :provider_without_slot do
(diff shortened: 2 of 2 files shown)


felixrieseberg:
LGTM 👍

felixrieseberg:
@BanzaiMan We at Microsoft would really appreciate a merge, currently pushes are broken for Azure Web Apps because we might not always have a master branch on the remote.

BanzaiMan:
Sorry about the delay. Please add specs, and I'll merge this.

Thank you!

kenhowardpdx:
@felixrieseberg & @A92hm One thing I've noticed w/ the azure provider is that it is copying all files in the build folder including `node_modules`. I'd suggest removing the `--force` flag.

In addition I've created a script for `before_deploy` that alters my `.gitignore` file. I have no use for TypeScript files in production and no use for JavaScript files during development. Removing the `--force` flag requires this change... but in the end I would prefer the `node_modules` to be installed via `npm install --production` on Azure during deploy.

kenhowardpdx:
One more thing worth mentioning. I'm deploying a branch that is not `master`. This depends on `master` existing in `refs/heads`. I'm not sure of the way around this... other than capturing the current branch name then `context.shell "git push --force --quiet #{git_target} HEAD:refs/heads/#{current_branch}`... thoughts?

A92hm:
Sorry for the delay everyone. I completely forgot about this.

@kenhowardpdx: I think this is a design strategy, I would suggest opening an issue for it so we can all discuss it.

@BanzaiMan I've added Spec for azure web apps. Let me know if you want me to change anything!

kenhowardpdx:
@A92hm Actually I misunderstood the command. Disregard my previous comments. :)

A92hm:
@kenhowardpdx No worries. I understand what you are getting at, and I've felt the need for a more smart deployment script.

As I use this more, I might end up making more changes that makes sense :)

putuyoga:
Hi, it seems i can not deploy from travis build to my new azure web app, same with a92hm, any idea ?

A92hm:
@putuyoga I've fixed the issue in this PR, but It needs to be merged in first.
@BanzaiMan I would appreciate if you could merge this.

BanzaiMan:
Has this been tested?

BanzaiMan:
Sorry about the delay.

A92hm:
No worries.

Yes, the original deployment was successful and had been tested. You just wanted us to provide specs.

I did another deployment right now just to be sure.
https://travis-ci.org/A92hm/travis-azure --> http://travis-test.azurewebsites.net/

PR #451, opened 2016-05-24
Update link to OpsWorks custom JSON line

joecorcoran:
Line in readme was linking to the wrong place.


BanzaiMan:
This is stale now. The deep linking to code really doesn't work, but I'm not sure how we can do better.

PR #457, opened 2016-05-31
Add AWS S3 delete option

5c077yP:
hey there,

this PR adds a `delete` option to the s3 deployment provider so that files, that were not synced get deleted. This option is set to false by default. 

My use case is the following. I have a webpack js application and for every release webpack creates a new file with some random string, so that there are no bad caches after a release. Without this delete option i would need to manually clean up old files. Currently i'm using the aws-cli which supports a `aws s3 sync --delete` to remove files that were not synced.

Let me know if I should provide any more information or updates to the code.

### DIFF
--- a/lib/dpl/provider.rb
+++ b/lib/dpl/provider.rb
@@ -1,6 +1,7 @@
 require 'dpl/error'
 require 'dpl/version'
 require 'fileutils'
+require 'json/pure'
 
 module DPL
   class Provider


icco:
@5c077yP you'll have to rebase or merge in master to get your code to pass.

5c077yP:
@icco rebased it 😄

PR #459, opened 2016-06-03
Add tests for ruby 2.3

icco:


icco:
@BanzaiMan PTAL

icco:
P.S. filed https://github.com/rvm/rvm/issues/3697 to try and get things fixed for the 2.3 failure.

BanzaiMan:
The current `rvm` is old and unaware of `2.3`. If you specify `2.3.0` it should work.

icco:
@BanzaiMan from what I can tell it's because binaries for your OS (Precise) and any version of ruby 2.3 don't exist, not because rvm doesn't know about 2.3. If there wasn't the `--binary`, I think it would just work... but maybe I'm misunderstanding.

BanzaiMan:
I'm closing this.

PR #479, opened 2016-07-12
Deis provider: unthrow error on push_app

Natsuke:
Hi,

We ran into some issue when using dpl with the deis provider. In some case the build triggered by deis failed but the error was not caught by dpl.

This fix remove the piped script on the `git push` and instead check the returned status code.

### DIFF
--- a/lib/dpl/provider/deis.rb
+++ b/lib/dpl/provider/deis.rb
@@ -81,7 +81,7 @@ module DPL
       end
 
       def push_app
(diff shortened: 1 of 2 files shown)


mattk42:
Removing these pipes will cause the output from the plugin to become unreadable. Could we just move to using something like the $PIPESTATUS variable to check the output of the git push instead of the formatting commands?

Natsuke:
Sounds good. I'll go this way and keep you in touch !

Fridus:
👍

PR #482, opened 2016-07-20
Move `json_pure` to on-demand dependency

BanzaiMan:


scott-christopher:
We're seeing `require: cannot load such file -- json/pure (LoadError)` errors with S3 deployments too. Should this perhaps be extended to include the `json_pure` gem in the `sss` group too?

PR #483, opened 2016-07-22
Remove json/pure from provider.rb

BanzaiMan:


BanzaiMan:
Also load `json/pure` early in `spec_helper.rb` so that we can avoid [TypeError](https://github.com/flori/json/issues/186).

PR #440, opened 2016-05-02
Support Deis Workflow (V2) client

mattk42:
This change adds support to install the Deis Workflow V2 client. The usage of the client has remained the same as in V1, but it is now installed using a different script.

Currently V2 is in beta and the client install script only supports installing the latests release of the V2 client. So for now this will install the latest V2 client if the "cli_version" option is set to 2. This will likely change once the installation script is out of beta and supports installing other clients.

I made no change to the tests, as they do not utilize the actual client and simply return a true for the shell out that is called.

This change has been tested against a working Deis V2 environment and the deploy succeeds as expected.

PR #491, opened 2016-08-09
Deis provider needs to execute push in bash

mattk42:
#479 wasn't working as expected. PIPESTATUS is a bash feature and the context.shell is a simple sh session. The result of this would be that all deploys would get marked as failed. 

To fix this I am executing the deployment in a bash session.

### DIFF
--- a/lib/dpl/provider/deis.rb
+++ b/lib/dpl/provider/deis.rb
@@ -81,7 +81,7 @@ module DPL
       end
 
       def push_app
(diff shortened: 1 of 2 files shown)


Joshua-Anderson:
@BanzaiMan is there anything that needs to happen for this to be merged?

BanzaiMan:
Sorry about the delay. I was traveling a bit.

BanzaiMan:
Has this been tested anywhere? If not, please rebase this beforehand. (I don't think it is a big deal, but the red text still scares me.)

mattk42:
I have been running from my own branch that has this and #440 merged together with no issues. I will go ahead and rebase.

PR #499, opened 2016-08-19
Fixes issue with `go get atlas-upload-cli` not working

solarce:
- Better handling of an empty `$GOPATH`
- the way `dpl` sets `GIT_HTTP_USER_AGENT` seems
  to cause issues with `go get` now, but removing it
  resolves the issue
- No longer any need to `make` the binary, `go get ...`
  installs and makes a working binary

### DIFF
--- a/lib/dpl/provider/atlas.rb
+++ b/lib/dpl/provider/atlas.rb
@@ -23,7 +23,6 @@ module DPL
           fi
           eval "$(gimme 1.6)" &> /dev/null
 
-          unset GIT_HTTP_USER_AGENT
           go get #{ATLAS_UPLOAD_CLI_GO_REMOTE}
           cp $HOME/gopath/bin/atlas-upload-cli $HOME/bin/atlas-upload
         fi
@@ -61,7 +60,9 @@ module DPL
       private
 
       def install_atlas_upload
-        context.shell ATLAS_UPLOAD_INSTALL_SCRIPT
+        without_git_http_user_agent do
+          context.shell ATLAS_UPLOAD_INSTALL_SCRIPT
+        end
       end
 
       def assert_app_present!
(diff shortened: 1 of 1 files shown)


solarce:
I don't think any of what I've changed affects the spec tests, but I'd appreciate your input @BanzaiMan

lib/dpl/provider/atlas.rb
```diff
@@ -16,14 +16,16 @@ class Atlas < Provider
...
+          fi
+          eval "$(gimme 1.6)" &> /dev/null
 
+          unset GIT_HTTP_USER_AGENT
           go get #{ATLAS_UPLOAD_CLI_GO_REMOTE}
-          pushd $HOME/gopath/src/#{ATLAS_UPLOAD_CLI_GO_REMOTE} &>/dev/null
-          make &>/dev/null
-          cp bin/atlas-upload $HOME/bin/atlas-upload
-          popd &>/dev/null
+          cp $HOME/gopath/bin/atlas-upload-cli $HOME/bin/atlas-upload
```
BanzaiMan: Is `go get` supposed to build the binary? I didn't notice that.

lib/dpl/provider/atlas.rb
```diff
@@ -16,14 +16,16 @@ class Atlas < Provider
...
-          export GOPATH="$HOME/gopath:$GOPATH"
-          eval "$(gimme 1.4.2)" &>/dev/null
+          if [ -z $GOPATH ]; then
+            export GOPATH="$HOME/gopath"
+          else
+            export GOPATH="$HOME/gopath:$GOPATH"
+          fi
+          eval "$(gimme 1.6)" &> /dev/null
 
+          unset GIT_HTTP_USER_AGENT
```
BanzaiMan: Is it safe to unset it here? I _think_ it is, but do keep in mind that another provider might depend on this value later. (I don't think that this `unset` does affects the `dpl` calls later.)

lib/dpl/provider/atlas.rb
solarce: It goes build it, if you run `go get -x ...` you'll see it in the output, https://travis-ci.org/travis-infrastructure/container-tests-org/builds/153417345#L202-L203 deploys successfully after `line 28` does the `cp`

lib/dpl/provider/atlas.rb
solarce: I honestly don't know, I didn't really look at how the shell handling works to determine if all deploy providers run within the same shell environment, but I could test it by adding another provider to my test repo

PR #504, opened 2016-09-13
Update to firebase-tools v3

### LINKED ISSUE #503: Provider firebase should remplace -f by --project
With firebase dpl use firebase-tools with option`-f #{options[:project]}` (https://github.com/travis-ci/dpl/blob/master/lib/dpl/provider/firebase.rb#L20) and this throw the error: `error: unknown option '-f'`

Watching [firebase-tools commands](https://github.com/firebase/firebase-tools#commands) we should use `-P` or `--project`

jorgecasar:
fix #503

### DIFF
--- a/lib/dpl/provider/firebase.rb
+++ b/lib/dpl/provider/firebase.rb
@@ -1,7 +1,7 @@
 module DPL
   class Provider
     class Firebase < Provider
-      npm_g 'firebase-tools@^2.1', 'firebase'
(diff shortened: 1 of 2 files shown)


BanzaiMan:
Has this been tested?

jorgecasar:
Yes, I use it to deploy my projects. Here is a screenshot:

firebase-deploy

To know more about [firebase-tools](https://github.com/firebase/firebase-tools)

PR #516, opened 2016-09-30
Unshallow before Heroku git deployment

BanzaiMan:
Heroku rejects pushes from a shallow clone like ours since Heroku
may not have all the objects necessary to build the application
reliably.


BanzaiMan:
Tested here: https://travis-ci.org/travis-ci/docs-travis-ci-com/builds/163896891

BanzaiMan:
https://travis-ci.org/travis-ci/docs-travis-ci-com/builds/163901071 shows that it also works with older (now-default) version of git.

PR #525, opened 2016-11-02
[PyPI] Add option to skip uploading documentation

lukeyeager:
*Fix https://github.com/travis-ci/dpl/issues/334, close https://github.com/travis-ci/dpl/pull/349*

This is a rebase and slight refactor of work from @eric-takealot in https://github.com/travis-ci/dpl/pull/349 after a merge conflict from https://github.com/travis-ci/dpl/pull/396.

> Has this been tested?
@BanzaiMan https://github.com/travis-ci/dpl/pull/349#issuecomment-149536356

With `skip_upload_docs: true`: https://travis-ci.org/lukeyeager/DIGITS/jobs/172784593
With `skip_upload_docs: false`: https://travis-ci.org/lukeyeager/DIGITS/jobs/172785158

PR #526, opened 2016-11-06
Updated the default PyPI upload URL

agronholm:
The "twine" tool uses this URL by default. The only reason why Travis still uses the legacy URL is because it's overridden in ~/.pypirc by this code. This new URL points to the newer Warehouse server which is the replacement for the old Cheeseshop PyPI implementation. Uploading to Warehouse adds the distribution to both servers (unlike the legacy URL which sadly does not seem to affect Warehouse), and has proven more reliable than the older server.

### DIFF
--- a/lib/dpl/provider/pypi.rb
+++ b/lib/dpl/provider/pypi.rb
@@ -1,7 +1,7 @@
 module DPL
   class Provider
     class PyPI < Provider
-      DEFAULT_SERVER = 'https://pypi.python.org/pypi'
+      DEFAULT_SERVER = 'https://upload.pypi.org/legacy/'
       PYPIRC_FILE = '~/.pypirc'
 
       def pypi_user


lib/dpl/provider/pypi.rb
```diff
@@ -1,7 +1,7 @@
 module DPL
   class Provider
     class PyPI < Provider
-      DEFAULT_SERVER = 'https://pypi.python.org/pypi'
+      DEFAULT_SERVER = 'https://upload.pypi.org/legacy/'
```
BanzaiMan: Fix the specs, please.

agronholm:
There's an additional wrinkle: the docs upload functionality. This is no longer supported on PyPI (Warehouse never supported it in the first place) and should be removed, but it may be beyond my current skills as I am not a Ruby developer.

lukeyeager:
@agronholm I have a pending fix for that: https://github.com/travis-ci/dpl/pull/525

agronholm:
@lukeyeager as I understand, your patch adds an option to skip docs upload. But the functionality will be *removed entirely* once Warehouse replaces Cheeseshop as the PyPI implementation. Therefore it makes more sense to remove this feature entirely.

lukeyeager:
Sure, that's fine. But my PR still enables users who don't want to upload docs in the short-term, and doesn't block anything long-term.

agronholm:
In any case, the necessary changes are more involved than my understanding of Ruby permits.

BanzaiMan:
See https://github.com/travis-ci/dpl/pull/659.

Issue #519, opened 2016-10-14
S3 upload should not set an empty content-type


stof:
When the mime type is not known by the `mime-types` gem, it would be better to set a `application/octet-stream` mime type than an empty string.

BanzaiMan:
Thanks for opening the issue. What are effects of not setting the `Content-Type` header?

stof:
if you don't set it when uploading, I think S3 automatically sets it to ``application/octet-stream`` (which means that browsers will download the file, while an empty content-type gets treated like a text content at least in Firefox and so displayed on screen even when it is not textual)

BanzaiMan:
Please test it with:

```yaml
deploy:
  provider: s3
  edge:
    branch: 's3-content-type'
  ⋮ # rest
```

stof:
Actually, it does not set it to ``application/octet-stream`` (this may be a feature of s3cmd for unknown types).

But omitting the content-type makes the file be downloaded by Firefox instead of being displayed as text as previously (which was very confusing when the file was not a text file). So I would vote +1 on your fix.

BanzaiMan:
@stof @sylveon @peternewman You all have expressed interest in resolving this issue. Could you test the PR one last time, so that we can merge and release the fix?

Thanks.

peternewman:
Not guilty sorry @BanzaiMan aside from pointing out it's a duplicate to try and keep the issue list clean.

BanzaiMan:
@peternewman Sorry for the noise!

PR #529, opened 2016-11-09
Skip setting Content-Type header if unknown

BanzaiMan:
If MIME::Types can't figure out the type, skip setting the Content-Type
header altogether.
When downloading, Amazon should specify 'application/octet-stream'.

Resolves https://github.com/travis-ci/dpl/issues/519


svenfuchs:
@BanzaiMan `master` (dpl v2) now has the same behavior as suggested by this PR (`compact` removes keys that have `nil` values) https://github.com/travis-ci/dpl/blob/master/lib/dpl/providers/s3.rb#L103

however, @sylveon, you are suggesting that we should instead send `application/octet-stream` as a default, instead?

svenfuchs:
I'm going to close this PR here though. Let's open a new issue if we think we should send a content type `application/octet-stream` by default.

svenfuchs:
After some further discussion we have come to the conclusion it would be best to use `application/octet-stream` as a default content type as you suggested @sylveon 

See https://github.com/travis-ci/dpl/commit/5c5149fe57de81cfa7c0b3343a5f6c3c5edc74b4

PR #534, opened 2016-11-23
Set remote to unique value in Deis provider

jschneiderhan:
I ran into an issue while trying to setup two deis deploys in my `.travis.yml` file. The same remote name ("deis") is currently used for both, which causes the second to error with the following message:

```
Error: Remote deis already exists, please run 'deis git:remote -f' to overwrite
```

This PR sets the remote name to the unique combo of controller-app. Some other implementation ideas include:

- allow the user to set a `remote` options -  this doesn't seem necessary if we can generate a unique name, nor do I see it being useful for any other situation
- specify the `-f` option to the `git:remote` command

I ended up going with creating a separate remote for each deploy instead of using `-f` because it mimics the way we perform manual deploys. If anyone has strong opinions on other implementation ideas, please let me know.

### DIFF
--- a/lib/dpl/provider/deis.rb
+++ b/lib/dpl/provider/deis.rb
@@ -2,8 +2,6 @@ module DPL
   class Provider
     class Deis < Provider
       
-      requires 'git'
-
       def install_deploy_dependencies
         install_url = determine_install_url
         context.shell "curl -sSL #{install_url} | bash -x -s #{option(:cli_version)}"
@@ -59,10 +57,6 @@ module DPL
         chmod(0740, path)
         context.env['GIT_SSH'] = path
 
-        unless context.shell "./deis git:remote --app=#{option(:app)}"
-          error 'Adding git remote failed.'
-        end
-
         wait_for_git_access
       end
 
@@ -71,9 +65,7 @@ module DPL
         max_retries=30
 
         #Get the deis git remote host and port
-        git=Git.open("./")
-        git_remote=git.remote("deis").url
-        remote_uri=git_remote.split("ssh://")[1].split("/")[0]
+        remote_uri=repository_url.split("ssh://")[1].split("/")[0]
         remote_host, remote_port = remote_uri.split(":")
         puts "Git remote is #{remote_host} at port #{remote_port}"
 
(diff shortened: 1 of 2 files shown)


Joshua-Anderson:
The -f flag is new, so it wouldn't work with older Deis versions. I think this is probably a problem with other git based deployments as well. Should we maybe try to add some code to all providers to delete the remote if it exists before attempting to add one.

jschneiderhan:
@Joshua-Anderson agreed that this probably affects other git based deployments as well. What did you have in mind for the code? Were you thinking we should update each provider individually, or come up with some consistent way to purge remotes before each deploy?

In a quick search, the only other provider I saw explicitly managing remotes was https://github.com/travis-ci/dpl/blob/424c7a61c341f184b8aa66df178d4aca31455a6f/lib/dpl/provider/scalingo.rb#L88. Everything else, if it is using git, must be hiding the remote management behind library cals, so it may be hard to determine which remote to remove without relying on those same libraries.

Joshua-Anderson:
@jschneiderhan Ah, my bad. I hadn't looked at the code recently and mis-remembered it. Most providers just push directly to the repository without adding a remote. I think that's probably the best approach. Just `git push ssh://git@deis-builder.<domain>.com:2222/<app>.git`

jschneiderhan:
@Joshua-Anderson I just tried a quick PoC removing the `git:remote` command and using an explicit repo URL in the push command. The push and deis deployment worked, but my travis deploy failed when trying to run a command. It looks like the `run` functionality requires the remote to be in place :(

The quick PoC code is [here](https://github.com/travis-ci/dpl/compare/master...meyouhealth:remote-explicit-push?expand=1) if you're curious.

Joshua-Anderson:
@jschneiderhan Without a remote the app need to be explicitly declared: `./deis run -a appname --command`

jschneiderhan:
@Joshua-Anderson ah yes, of course! The `run` command uses the controller API, not the builder.

I've updated the PR to use an explicit repository url in the git push.

lib/dpl/provider/deis.rb
```diff
@@ -95,14 +87,28 @@ def remove_key
         end
       end
 
+      def repository_url
+        "ssh://git@#{builder_hostname}:2222/#{option(:app)}.git"
+      end
+
+      def builder_hostname
```
BanzaiMan: Does this naming always work? I mean, `-builder` is now required in all cases.

Joshua-Anderson: Yes, the builder component has a static hostname.

BanzaiMan:
OK. Do we have a test deployment?

jschneiderhan:
I've been using my PR branch for continuous deployments for a few weeks without any issues, if that's what you mean. If you're wondering about some sort of acceptance test beyond the `deis_spec` then no, but I'm not sure how to create one either.

BanzaiMan:
@jschneiderhan Yes, I mean test deployment. Do you have build log URLs to share? If you'd rather not discuss them in public, please email us at support@travis-ci.com. Thanks.

jschneiderhan:
@BanzaiMan I just sent a couple of links to private builds to support@travis-ci.com.

PR #537, opened 2016-12-06
Relax constraint on chef version

kamaradclimber:
Since dpl now runs with ruby >=2.2.0, constraining chef to <12.0 is not
necessary anymore.

This reverts commit ba764dddfedbae3128ff4941641094eae6718fc4.


BanzaiMan:
Thanks for taking the time to open this PR. Unfortunately, we cannot accept this, since the deployment on our platform still happens with Ruby 1.9.3.

PR #541, opened 2016-12-10
packagecloud client bump to 1.0.5

capotej:
Bumps the [`packagecloud-ruby`](https://github.com/computology/packagecloud-ruby) client in the [`packagecloud`](https://github.com/travis-ci/dpl#packagecloud) provider to [`v1.0.5`](https://rubygems.org/gems/packagecloud-ruby/versions/1.0.5) which fixes several bugs and adds python support.

#### [computology/python-test-packages](https://github.com/computology/python-test-packages)
https://github.com/computology/python-test-packages/blob/master/.travis.yml#L20
https://travis-ci.org/computology/python-test-packages#L1348

#### [capotej/http](https://github.com/capotej/http)
https://travis-ci.org/capotej/http/builds/182718448
https://github.com/capotej/http/blob/master/.travis.yml#L34

Not sure if any authentication/verification is needed, but [I work for packagecloud.io](https://github.com/orgs/computology/people).

Let me know if you need anything else to merge this, thanks!

### DIFF
--- a/lib/dpl/provider/packagecloud.rb
+++ b/lib/dpl/provider/packagecloud.rb
@@ -65,7 +65,7 @@ module DPL
         source_files = {}
         glob_args = ["**/*"]
         package = ::Packagecloud::Package.new(file: orig_filename)
-        result = @client.package_contents(@repo, package)
+        result = @client.package_contents(@repo, package, get_distro(@dist))
         if result.succeeded
           package_contents_files = result.response["files"].map { |x| x["filename"] }
           Dir.chdir(options.fetch(:local_dir, Dir.pwd)) do
--- a/lib/dpl/provider/packagecloud.rb
+++ b/lib/dpl/provider/packagecloud.rb
@@ -64,7 +64,7 @@ module DPL
       def get_source_files_for(orig_filename)
         source_files = {}
         glob_args = ["**/*"]
-        package = ::Packagecloud::Package.new(file: orig_filename)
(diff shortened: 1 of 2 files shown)


lib/dpl/provider/packagecloud.rb
```diff
@@ -64,8 +64,8 @@ def is_source_package?(filename)
       def get_source_files_for(orig_filename)
         source_files = {}
         glob_args = ["**/*"]
-        package = ::Packagecloud::Package.new(open(orig_filename))
-        result = @client.package_contents(@repo, package)
+        package = ::Packagecloud::Package.new(file: orig_filename)
```
BanzaiMan: Please refrain from using keyword arguments. Our deployment runs on 1.9.3 right now, and I think this will break it.

capotej:
Good catch, fixed.

BanzaiMan:
Am I correct to assume that the call signature for `::Packagecloud::Package.new` changed?

capotej:
Yes, it used to take an IO object, now it takes an option hash and can use string paths or IO objects: https://github.com/computology/packagecloud-ruby/commit/6e647d5eb770ca4c0bc6aef14607ffbfbb4d2ed4#diff-7372dd819be8b828820934318bab0fca

capotej:
It is worth noting however, that this breaking change is an internal detail encapsulated by the packagecloud provider and end-users of `dpl` will not be affected.

capotej:
Thanks!

PR #549, opened 2016-12-15
Fix GAE deployment version error in gcloud sdk v138.0.0.

M157q:
In gcloud sdk v138.0.0,
gcloud app deploy with args `--version ""` will get error below:

ERROR: (gcloud.app.deploy) argument --version/-v: Bad value []: May only
contain lowercase letters, digits, and hyphens. Must begin and end with
a letter or digit. Must not exceed 63 characters.

So, if user didn't set the version and we want to use the gcloud
default generated version in this deployment, we just don't add
`--version`.

### DIFF
--- a/lib/dpl/provider/gae.rb
+++ b/lib/dpl/provider/gae.rb
@@ -47,7 +47,7 @@ module DPL
       end
 
       def version
-        options[:version] || ''
+        options[:version]
       end
 
       def config
@@ -72,7 +72,7 @@ module DPL
         command << " --verbosity \"#{verbosity}\""
         command << " --project \"#{project}\""
         command << " app deploy \"#{config}\""
-        command << " --version \"#{version}\""
+        command << (version ? " --version \"#{version}\"" : '')
         command << " --#{no_promote ? 'no-' : ''}promote"
         command << (no_stop_previous_version ? ' --no-stop-previous-version' : '')
         unless context.shell(command)
--- a/spec/provider/gae_spec.rb
+++ b/spec/provider/gae_spec.rb
@@ -8,7 +8,7 @@ describe DPL::Provider::GAE do
 
   describe '#push_app' do
     example 'with defaults' do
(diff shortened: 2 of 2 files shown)


M157q:
Got error when used `--version=""`


---

No error after removed `--version=""`

M157q:
+ [Failed Travis job before using this patch](https://travis-ci.com/Tagtoo/api/jobs/59571014)
+ [Passed Travis job after using this patch](https://travis-ci.com/Tagtoo/api/builds/36716118)

lib/dpl/provider/gae.rb
```diff
@@ -72,7 +72,7 @@ def push_app
         command << " --verbosity \"#{verbosity}\""
         command << " --project \"#{project}\""
         command << " app deploy \"#{config}\""
-        command << " --version \"#{version}\""
+        command << (version ? " --version \"#{version}\"" : '')
```
BanzaiMan: I don't think this works if `version` was explicitly given as `""`, as unlikely as it might be. (This actually is relevant to the `no_stop_previous_version` option, now that I think about it.)

A more robust way for this is:

```ruby
command << " --version \"#{version}\"" unless version.empty?
```

(And similar for `no_stop_previous_version`.)

BanzaiMan: I'll fix this after the merge.

M157q:
@BanzaiMan I got this error after using your patch.


---

Also, Travis build of this commit https://github.com/travis-ci/dpl/commit/30a3440e416bbea96fec6f9335e13745f5d94a75 failed
https://travis-ci.org/travis-ci/dpl/builds/184266319

BanzaiMan:
Sorry about that. That's fixed already.

M157q:
Change back to 

```
      def version
        options[:version] || ''
      end
```

and

```
      def version
        options[:no_stop_previous_version] || ''
      end
```

should fix this error, I guess?

BanzaiMan:
The release 1.8.26 should work. See https://github.com/travis-ci/dpl/commit/719220ebf399e040bfb9bbfebedb7c3dc0064209.

M157q:
Oh, cool.
I didn't see that commit.
Actually, I am not familiar with Ruby.
Thanks!

PR #563, opened 2017-01-05
Invoke GAE deployment commands with Python 2.7

BanzaiMan:
Resolves https://github.com/travis-ci/travis-ci/issues/6913.

This works by spawning a new bash process, activate Python 2.7, and execute the commands. See the issue above for tests.

### DIFF
--- a/lib/dpl/provider/gae.rb
+++ b/lib/dpl/provider/gae.rb
@@ -11,7 +11,7 @@ module DPL
       GCLOUD="#{INSTALL}/#{NAME}/bin/gcloud"
 
       def with_python_2_7(cmd)
-        "source #{context.env['HOME']}/virtualenv/python2.7/bin/activate; #{cmd}"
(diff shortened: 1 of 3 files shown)


lib/dpl/provider/gae.rb
```diff
@@ -10,6 +10,10 @@ class GAE < Provider
       BOOTSTRAP="#{INSTALL}/#{NAME}/bin/bootstrapping/install.py"
       GCLOUD="#{INSTALL}/#{NAME}/bin/gcloud"
 
+      def with_python_2_7(cmd)
+        context.shell("bash -c 'source #{context.env['HOME']}/virtualenv/python2.7/bin/activate; #{cmd}'")
```
demosdemon: 1. Since you're invoking bash, I don't think it's necessary to expand `~` yourself anymore.
2. I wonder if you should escape `cmd`. There's no cases now where it's invoked with an apostrophe, but in the future...

PR #569, opened 2017-01-13
Remove "experimental" tags from all providers

BanzaiMan:
This removes all "experimental" warnings.

These providers have gone through usage in the wild and most obvious issues have been addressed.

### DIFF
--- a/lib/dpl/provider/atlas.rb
+++ b/lib/dpl/provider/atlas.rb
@@ -28,8 +28,6 @@ module DPL
         fi
       EOF
 
-      experimental 'Atlas'
-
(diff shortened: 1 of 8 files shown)


BanzaiMan:
The current failure is due to `ffi` 1.9.16 being broken with Ruby 1.9.3.

BanzaiMan:
`ffi` 1.9.17 was released with the fix for Ruby 1.9.3.

svenfuchs:
I think we can close this. In v2 we now have a maturity status model using `dev|alpha|beta|stable`.

PR #586, opened 2017-02-14
Ansible deploy type

adamkingit:
Create a new provider that installs Ansible, and runs the playbook. Playbook options are represented in the YAML as described in README.md

Quiet the pip install so only error and critical messages are displayed.

Example successful deployment using the ansible type can be found at: 
https://travis-ci.org/adamkingit/echo-vcap/builds/201333182


webknjaz:
Nice, but is there a simple way to setup auth @ remote ansible deployment host?

BanzaiMan:
Sorry for the long delay. If you are still interested in this feature (I hope you are!), please update the code (for dpl 1.9.x) and open a new PR.

Thanks!

PR #591, opened 2017-02-23
raises error when cf push failed

### LINKED ISSUE #441: Cloud Foundry Deployment exits with status code 0 even if cf push fails
If the `cf push` step fails, the exit code should be 1. Currently, this is causing my builds to be marked a success, even if the application fails to deploy.

shawnzhu:
fixes #441

### DIFF
--- a/lib/dpl/provider/cloud_foundry.rb
+++ b/lib/dpl/provider/cloud_foundry.rb
@@ -23,7 +23,9 @@ module DPL
       end
 
       def push_app
-        context.shell "./cf push#{manifest}"
+        error 'Failed to push app' unless context.shell("./cf push#{manifest}")
+
+      ensure
         context.shell "./cf logout"
       end
 
--- a/spec/provider/cloudfoundry_spec.rb
+++ b/spec/provider/cloudfoundry_spec.rb
@@ -43,6 +43,10 @@ describe DPL::Provider::CloudFoundry do
   end
 
   describe "#push_app" do
+    before do
+      allow(provider.context).to receive(:shell).and_return(true)
+    end
+
     example "With manifest" do
       expect(provider.context).to receive(:shell).with('./cf push -f worker-manifest.yml')
       expect(provider.context).to receive(:shell).with('./cf logout')
@@ -54,6 +58,14 @@ describe DPL::Provider::CloudFoundry do
       expect(provider.context).to receive(:shell).with('./cf push')
       expect(provider.context).to receive(:shell).with('./cf logout')
       provider.push_app
+    end
+
+    example 'Failed to push' do
+      allow(provider.context).to receive(:shell).and_return(false)
+
+      expect(provider.context).to receive(:shell).with('./cf push -f worker-manifest.yml')
+      expect(provider.context).to receive(:shell).with('./cf logout')
+      expect{provider.push_app}.to raise_error(DPL::Error, 'Failed to push app')
     end
   end
 end


jossy:
Getting this merged would be super helpful! I can't see how Travis' Cloud Foundry deployment can be used for non-trivial projects if a build does not fail on push error.

afeld:
Hmm, seems that `context.shell` won't throw an error if a command fails, meaning that there would need to be a corresponding change everywhere in the codebase `context.shell` is invoked. Thinking it would be much simpler to have `context.shell` throw an error when the underlying `system` call returns `false`, rather than putting the onus on the callers to check for it.

umaar:
Right now, a if a Cloud Foundry deployment fails, even it returns with a failing exit code, this is not picked up by Travis and as such, the Travis build is marked as successful even the deployment failed. Could this PR be merged?

lib/dpl/provider/cloud_foundry.rb
```diff
@@ -23,8 +23,8 @@ def needs_key?
       end
 
       def push_app
-        context.shell "./cf push#{manifest}"
-        context.shell "./cf logout"
+        error 'Failed to push app' if context.shell("./cf push#{manifest}") == false
```
BanzaiMan: `if x == false` is cumbersome.

```ruby
error 'Failed to push app' unless context.shell("./cf push#{manifest}")
```

lib/dpl/provider/cloud_foundry.rb
```diff
@@ -23,8 +23,8 @@ def needs_key?
...
+        ensure context.shell "./cf logout"
```
BanzaiMan: `ensure` block should be introduced here.

```ruby
⋮
ensure
  context.shell "./cf logout"
⋮
```

lib/dpl/provider/cloud_foundry.rb
```diff
@@ -23,8 +23,10 @@ def needs_key?
       end
 
       def push_app
-        context.shell "./cf push#{manifest}"
-        context.shell "./cf logout"
+        error 'Failed to push app' unless context.shell("./cf push#{manifest}")
```
BanzaiMan: This `context.shell` call appears to return `false` in specs. Could you look into this?

lib/dpl/provider/cloud_foundry.rb
```diff
@@ -23,8 +23,10 @@ def needs_key?
...
+
+        ensure
```
BanzaiMan: Stylistically, this line and the line below should be one level up (delete 2 spaces at the beginning of the lines).

shawnzhu:
@BanzaiMan review comments addressed, would you please review it again?

BanzaiMan:
@shawnzhu Great! Do you have example builds that show how this would work (one for success, one for failure)?

shawnzhu:
@BanzaiMan no. I need to figure out how to show outcome of this change in travis-ci.org :)

what is the special configuration in `.travis.yml` to use my branch of `travis-ci/dpl`?

BanzaiMan:
@shawnzhu https://github.com/travis-ci/dpl/blob/master/TESTING.md

shawnzhu:
Here's one example:
* source: https://github.com/shawnzhu/get-started-python/tree/test-dpl
* [`.travis.yml`](https://github.com/shawnzhu/get-started-python/blob/test-dpl/.travis.yml)
* a failed deployment fails Travis build: https://travis-ci.org/shawnzhu/get-started-python/builds/230193532

jwickens:
Would be great to get this merged!

PR #597, opened 2017-03-12
Prerelease on branch

mariogrip:
This adds option to explicitly specify branches to identify as prerelease on GitHub.

example:

```bash
dpl --provider=releases --api-key=key --file="build.tar.gz" --repo=mariogrip/github-binary-release --prerelease=true prerelease_on_branch=master
```

example running on travis:
https://github.com/mariogrip/github-binary-release/releases
https://github.com/mariogrip/github-binary-release
https://travis-ci.org/mariogrip/github-binary-release


BanzaiMan:
Sorry for the delay.

So, the use case here is that you want to create a pre-release on a particular branch, right?

`prerelease: true` should already work, and if you use conditional deployment (either with `on.tags: true`, or by adding a deployment stage only on the desired branch), what you want to accomplish is achieved.

Introducing a new option seems too heavy-handed.

PR #617, opened 2017-04-21
Script: User-friendly error messages if not exec.

jakubholynet:
The script provider checks whether the script file exists and is executable and provides user-friendly error messages if not.

### DIFF
--- a/lib/dpl/provider/script.rb
+++ b/lib/dpl/provider/script.rb
@@ -1,3 +1,5 @@
+require 'pathname'
+
 module DPL
   class Provider
     class Script < Provider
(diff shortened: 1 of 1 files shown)


BanzaiMan:
I do not think this is a good idea. `script` can be a bare command; e.g.,

```yaml
deploy:
  provider: script
  script: echo hello
```

and this will break it.

```sh-session
$ ruby -r pathname -e 'puts Pathname.new("echo foo").file?'
false
```

I believe appropriate message should be printed when the script is actually invoked.

jakubholynet:
Good point! I havent noticed that use case. How to fix it then?
fre. 21. apr. 2017 kl. 17.59 skrev Hiro Asari <notifications@github.com>:

BanzaiMan:
I think we need to change the way we execute system commands.

Currently we are [using <tt>Kernel#system</tt>](https://github.com/travis-ci/dpl/blob/v1.8.31/lib/dpl/provider.rb#L85), but I think this is swallowing the exception the subshell is throwing (e.g., file not found, or permission denied). We are only displaying the exit status (in this case, code 127), which is not very helpful in deciphering what actually went wrong.

jakubholynet:
Exactly
lør. 22. apr. 2017 kl. 00.12 skrev Hiro Asari <notifications@github.com>:

BanzaiMan:
I'll look at such a change for 1.9.

In the mean time, I am closing this PR.

PR #618, opened 2017-04-24
Add packagecloud re-upload

waghanza:
Hello,

For some reason, I wanted to upload the same version of a package on [PackageCloud](https://packagecloud.io).

Currently, uploading the same version cause **deploy** phase to fail.

This `PR` add ability to order package **re**upload.

A successful usage could be found on https://travis-ci.org/waghanza/pkg-geoip2/builds/223850598


lib/dpl/provider/packagecloud.rb
```diff
@@ -88,6 +88,7 @@ def get_source_files_for(orig_filename)
       end
 
       def push_app
+        forced = options.fetch(:force) { nil }
```
BanzaiMan: Since we don't use the key value in the block, it is simpler to say:

```ruby
forced = options.fetch(:force, nil)
```

In fact, since the absence of the `:force` key in the `options` hash results in `nil`, you can also write:

```ruby
forced = options[:force]
```

lib/dpl/provider/packagecloud.rb
```diff
@@ -109,6 +110,11 @@ def push_app
         end
 
         packages.each do |package|
+          if forced
+            log "Deleting package: #{package.filename}"
+            distro, distro_release = @dist.split("/")
+            @client.delete_package(@repo, distro, distro_release, package.filename)
```
BanzaiMan: What would happen if the deletion fails?

lib/dpl/provider/packagecloud.rb
waghanza: I prefer the firs solution, as clearer

lib/dpl/provider/packagecloud.rb
waghanza: you're success / failure are not treated 👍

slodki:
Will it work with auto-rename "feature" of packagecloud?

Scenario:
- [successfully deployed](https://travis-ci.org/slodki/moneymanagerex/jobs/229979491#L3545) with Travis CI package `ubuntu-zesty-mmex-1.4.0-ALPHA.deb`
- package is [presented and ready to download](https://packagecloud.io/app/slodki/mmex_dev/search?dist=ubuntu%2Fzesty) as `mmex_1.4.0_amd64.deb`
- I get `filename: not found` [error](https://travis-ci.org/slodki/moneymanagerex/builds/230017799#L361) when I try do delete it using `package_cloud yank slodki/mmex_dev/ubuntu/zesty ubuntu-zesty-mmex-1.4.0-ALPHA.deb`

It's impossible to delete uploaded file using original filename. How can I calculate correct package name for deletion? I cannot find any info about this "feature" at packagecloud.

Have you tested your proposal with this scenario?

waghanza:
@slodki in fact, this feature is a simple as _delete before upload_ so it's kinda **rename**

BanzaiMan:
Documentation update PR, too, please.

waghanza:
@BanzaiMan thanks for merging. What do you need more as documentation ?

BanzaiMan:
There is a new option. That should be documented. https://docs.travis-ci.com/user/deployment/packagecloud

PR #629, opened 2017-05-13
Fix elastic beanstalk bucket creation

ef4:
This was calling a nonexistent method.

Perhaps everyone has been pre-creating their buckets, in which case they wouldn't hit this bug.

### DIFF
--- a/lib/dpl/provider/elastic_beanstalk.rb
+++ b/lib/dpl/provider/elastic_beanstalk.rb
@@ -103,7 +103,7 @@ module DPL
       end
 
       def create_bucket
-        s3.buckets.create(bucket_name)
+        s3.create_bucket(bucket: bucket_name)
       end
 
       def files_to_pack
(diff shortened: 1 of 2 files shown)


lib/dpl/provider/elastic_beanstalk.rb
```diff
@@ -103,7 +103,7 @@ def bucket_exists?
       end
 
       def create_bucket
-        s3.buckets.create(bucket_name)
+        s3.create_bucket({ bucket: bucket_name })
```
BanzaiMan:
```ruby
s3.create_bucket(bucket: bucket_name)
```

I feel that it looks better without the brace here.

ef4: Haha, I agree. That's what happens when you spend years writing Javascript and haven't done Ruby in a long time. I will update the PR.

ef4:
I updated the test suite to show the fix. It's all mocked of course, so it's not a high-fidelity test for compatibility with aws-sdk, but it's consistent with how the existing tests were already written.

In the test, I also changed:

> expect(provider).to receive(:s3).and_return(s3_mock).twice

to

> allow(provider).to receive(:s3).and_return(s3_mock)

Because I really don't think the number of calls to the `s3` method is relevant to the behavior under test. It's probably worth refactoring that out of all the other tests too and making it part of the setup.

BanzaiMan:
I'm closing this in favor of https://github.com/travis-ci/dpl/pull/638.

PR #631, opened 2017-05-16
[WIP] Log better, and exit with consistent command status

BanzaiMan:
The idea is to call the system command while capturing STDERR
and STDOUT.
Also capture the "shell" commands' exit status, and use it to
exit the `dpl` command itself, so that nonzero deployment could
be used to signify deployment failure, which will mark the build
a failure as well.


BanzaiMan:
Test 1: https://travis-ci.org/BanzaiMan/travis_production_test/builds/232952041#L219

PR #638, opened 2017-05-24
Use AWS SDK v2 call to create S3 bucket in Elastic Beanstalk

BanzaiMan:


BanzaiMan:
Tested: https://travis-ci.org/BanzaiMan/travis_production_test/builds/235798141#L241

It is failing to upload the file, but that is because my repo's setup is incomplete. The bucket is actually created, as shown here:

s3_management_console

PR #645, opened 2017-06-03
Prep Heroku deployment providers for V3

BanzaiMan:
This PR prepares Heroku deployment providers for the upcoming switch to V3.

Currently, only the `api` strategy uses the API V3 to deploy, but this PR expands it into `git`.

Here is what happens with each provider
* The `anvil` strategy is no longer supported; the endpoint the provider talks to no longer works, so this is not an issue.
* The API V3 calls using Faraday is moved from the `api` provider to `generic`, which will then be used with `git` provider.
* The `git-ssh-key` strategy will be shut down on June 26th, 2017. We add a deprecation warning. In order to keep this provider working until the sunset while moving `git` (which is the parent class of `git-ssh-key`), we copy all the inherited methods from the legacy API calls.
* The `git-deploy-key` is not touched, but, given that it inherits from `git-ssh-key` and makes deprecated calls, it will also be shut down on June 26th.


BanzaiMan:
https://travis-ci.org/travis-ci/jwt-decoder-example/builds/238854015 shows deployments with

1. [`api`](https://travis-ci.org/travis-ci/jwt-decoder-example/builds/238854015#L199)
1. [`git`](https://travis-ci.org/travis-ci/jwt-decoder-example/builds/238854015#L319)
1. [`git-ssh-key`](https://travis-ci.org/travis-ci/jwt-decoder-example/builds/238854015#L429)

PR #651, opened 2017-06-07
Add support for message in Firebase provider

szeestraten:
Adds support for adding optional messages when deploying with the Firebase provider.

Please take a look at `spec/provider/firebase_spec.rb` as I'm not totally sure how the test should be.

Fixes travis-ci/travis-ci#7872


BanzaiMan:
Has this been tested?

I am particularly curious to know if it works when the message contains spaces.

szeestraten:
@BanzaiMan Didn't do my due diligence and test with spaces as you correctly spotted, sorry about that. Added a fix for that.

I did a test deploy which worked (see below), but I haven't used rspec so I need to read up on that if it's way off.

Test build/deployment
- https://travis-ci.org/travis-ci/dpl/builds/240910732
- [Screenshot from Firebase console](https://user-images.githubusercontent.com/614823/26948629-5f67a7ba-4c97-11e7-8003-de891a447378.png) with message

PR #659, opened 2017-06-30
Switch the default PyPI server to https://upload.pypi.org/legacy/

BanzaiMan:
The old 'https://pypi.python.org/pypi' will be shut down on
2017-07-03.

https://mail.python.org/pipermail/distutils-sig/2017-June/030766.html

See also https://github.com/travis-ci/dpl/issues/655


BanzaiMan:
https://github.com/travis-ci/dpl/issues/655#issuecomment-312300661 confirms that it works.

Issue #660, opened 2017-07-01
pypi's upload_docs crashes


akuzminsky:
when dpl uploads a python package to pypi it crashes on  `upload_docs` step
```
running upload_docs
Upload_docs command is deprecated. Use RTD instead.
creating /home/travis/build/twindb/proxysql-tools/build
creating /home/travis/build/twindb/proxysql-tools/build/sphinx
creating /home/travis/build/twindb/proxysql-tools/build/sphinx/doctrees
creating /home/travis/build/twindb/proxysql-tools/build/sphinx/html
Traceback (most recent call last):
  File "setup.py", line 54, in <module>
    tests_require=test_requirements
  File "/opt/python/2.7.13/lib/python2.7/distutils/core.py", line 151, in setup
    dist.run_commands()
  File "/opt/python/2.7.13/lib/python2.7/distutils/dist.py", line 953, in run_commands
    self.run_command(cmd)
  File "/opt/python/2.7.13/lib/python2.7/distutils/dist.py", line 971, in run_command
    cmd_obj.ensure_finalized()
  File "/opt/python/2.7.13/lib/python2.7/distutils/cmd.py", line 109, in ensure_finalized
    self.finalize_options()
  File "/home/travis/virtualenv/python2.7.13/lib/python2.7/site-packages/setuptools/command/upload_docs.py", line 65, in finalize_options
    self.target_dir = build_sphinx.builder_target_dir
  File "/opt/python/2.7.13/lib/python2.7/distutils/cmd.py", line 105, in __getattr__
    raise AttributeError, attr
AttributeError: builder_target_dir
```
full log https://travis-ci.org/twindb/proxysql-tools/jobs/248653364#L1124

BanzaiMan:
@dstufft May we have your advice, please?

dstufft:
PyPI deprecated and has removed the upload_docs API endpoint. You can see in the error message:

```
Upload_docs command is deprecated. Use RTD instead.
```

The suggested replacement is ReadTheDocs for these projects, but projects may of course choose elsewhere.

akuzminsky:
Yep, I use RTD. 
That error however prevented successful package upload. `skip_upload_docs: true` fixed that.

PR #668, opened 2017-07-12
Skip upload_docs by default for PyPI provider

### LINKED ISSUE #665: Change skip_upload_docs defaults to true
According to #660, `skip_upload_docs` is now actually required given that PyPI no longer supports uploading docs so the default should be `true`.

BanzaiMan:
https://github.com/travis-ci/dpl/issues/660 points to the change in default behavior. We should recognize this change, and change the default.

I think this is less risky than usual, since the default now leads to deployment failure.

This resolves #665.

### DIFF
--- a/README.md
+++ b/README.md
@@ -281,7 +281,7 @@ For authentication you can also use Travis CI secure environment variable:
 * **password**: PyPI Password.
 * **server**: Optional. Only required if you want to release to a different index. Follows the form of 'https://mypackageindex.com/index'. Defaults to 'https://pypi.python.org/pypi'.
 * **distributions**: Optional. A space-separated list of distributions to be uploaded to PyPI. Defaults to 'sdist'.
-* **skip_upload_docs**: Optional. When set to `true`, documentation is not uploaded. Defaults to `false`.
+* **skip_upload_docs**: Optional. When set to `false`, documentation is uploaded. Defaults to `true`.
 * **docs_dir**: Optional. A path to the directory to upload documentation from. Defaults to 'build/docs'
 
 #### Environment variables:
--- a/README.md
+++ b/README.md
@@ -279,9 +279,13 @@ For authentication you can also use Travis CI secure environment variable:
 
 * **user**: PyPI Username.
 * **password**: PyPI Password.
-* **server**: Optional. Only required if you want to release to a different index. Follows the form of 'https://mypackageindex.com/index'. Defaults to 'https://pypi.python.org/pypi'.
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -281,7 +281,7 @@ For authentication you can also use Travis CI secure environment variable:
 * **password**: PyPI Password.
 * **server**: Optional. Only required if you want to release to a different index. Follows the form of 'https://mypackageindex.com/index'. Defaults to 'https://pypi.python.org/pypi'.
 * **distributions**: Optional. A space-separated list of distributions to be uploaded to PyPI. Defaults to 'sdist'.
-* **skip_upload_docs**: Optional. When set to `true`, documentation is not uploaded. Defaults to `false`.
+* **skip_upload_docs**: Optional. When set to `false`, documentation is uploaded. Defaults to `true`.
```
nicoddemus: Perhaps it is worth mentioning that at the moment setting this to `false` will result in a build failure for PyPI? Otherwise someone might set it to `false` and file an error report.

BanzaiMan: Ah, right. Maybe we should remove this option altogether?

BanzaiMan: Or, is there an PyPI-compatible server that accepts the docs upload?

nicoddemus: > Ah, right. Maybe we should remove this option altogether?

If removing it won't break builds which currently set it, I think this is the way to go. Perhaps just leave t in the docs but mark it as "deprecated and no longer having any effect".

> Or, is there an PyPI-compatible server that accepts the docs upload?

readthedocs.com is now the recommended service for that, but it works differently by pulling directly from the repository.

README.md
```diff
@@ -279,9 +279,12 @@ For authentication you can also use Travis CI secure environment variable:
 
 * **user**: PyPI Username.
 * **password**: PyPI Password.
-* **server**: Optional. Only required if you want to release to a different index. Follows the form of 'https://mypackageindex.com/index'. Defaults to 'https://pypi.python.org/pypi'.
+* **server**: Optional. Only required if you want to release to a different index. Follows the form of 'https://mypackageindex.com/index'. Defaults to 'https://upload.pypi.org/legacy/'.
 * **distributions**: Optional. A space-separated list of distributions to be uploaded to PyPI. Defaults to 'sdist'.
-* **skip_upload_docs**: Optional. When set to `true`, documentation is not uploaded. Defaults to `false`.
+* **skip_upload_docs**: Optional. When set to `false`, documentation is uploaded. Defaults to `true`.
+  Note that upload.pypi.org does not support document uploading, so if you set this option to `false`,
+  your deployment fails, unless you specify the server that supports this option.
```
nicoddemus: Oh indeed, I missed this use case. 👍

plaindocs:
:+1:

PR #670, opened 2017-07-17
Truncate output of git commit in pages provider

iamEAP:
Prevents build/deploy errors when more than 10k files are `git commit`'d and the build log grows too large.

### DIFF
--- a/lib/dpl/provider/pages.rb
+++ b/lib/dpl/provider/pages.rb
@@ -59,7 +59,7 @@ module DPL
         context.shell "git config user.name '#{@gh_name}'"
         context.shell "echo '#{@gh_fqdn}' > CNAME" if @gh_fqdn
         context.shell 'git add .'
-        context.shell "git commit -m 'Deploy #{@project_name} to #{@gh_ref}:#{@target_branch}'"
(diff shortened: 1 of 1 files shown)


lib/dpl/provider/pages.rb
```diff
@@ -58,7 +58,7 @@ def github_deploy
         context.shell "git config user.email '#{@gh_email}'"
         context.shell "git config user.name '#{@gh_name}'"
         context.shell "echo '#{@gh_fqdn}' > CNAME" if @gh_fqdn
-        context.shell 'git add .'
+        context.shell 'git add . > /dev/null'
```
BanzaiMan: How about limiting the lines shown, instead of suppressing them all?

`| head` will show 10, and if there are more lines, you'd want to show ellipses for omission.

iamEAP: I like it. Perhaps something like...

```sh
git add . | tee >(head >&2) | [ $(wc -l) -gt 10 ] && echo "..."
```

BanzaiMan: Hmm. Does `git add .` show the files being added? I don't think it does, unless `-v` is given.

iamEAP: I think you're right. I think the problem I'm seeing is actually on the commit, e.g. https://travis-ci.org/drush-io/help/builds/254300569#L177

iamEAP:
@BanzaiMan changes updated to reflect discussion above.

Turns out process substitution isn't POSIX, so I head to do something less elegant.

Ran a sample build/deploy here with success: https://travis-ci.org/drush-io/help/builds/257417577#L291

PR #676, opened 2017-07-28
Fixing AWS Lambda push, adding all other SDK options

mboudreau:
Not sure if this works fully yet, it does passes the tests, but not tests for the new stuff.  Need to try it out first on my own projects to see if all options are correct.  Need to update documentation.


mboudreau:
@BanzaiMan I'm trying to get the parsing of options working.  The yaml I'm trying to parse is something like this:

```
provider: lambda
function_name: function
module_name: index
handler_name: handler
zip: package.zip
region: ap-southeast-2
runtime: nodejs6.10
timeout: 60
environment_variables:
  CLIENT_ID:
    secure: "something-secure"
  CLIENT_SECRET:
    secure: "something-secure"
  LOG_LEVEL: "debug"
```
But when I try to log/output the `:options` in the lambda provider, it seems to completely ignore `environment_variables` and not show anything.  I'm trying to keep it as a dictionary/map because I think it's simpler than using an array.  Is there a way to do this?  I can't seem to figure it out.  This is the only thing blocking me to make this PR 100% of what I'm trying to do, plus fix the tests and add some documentation.

Cheers.

iamkirkbater:
I'm really loving the work you're putting into this, I can only imagine how many people other than just me are enthusiastically waiting for this.  

I know you said you're trying to stay away from the array, but I think it might be a good solution here because in .travis.yml files for defining test builds env variables are defined as an array of key value pairs.  I think consistency is key.

I also don't know how the other deployment options are doing it, but I just wanted to throw in my $0.02.

Again, thanks for the work you're putting into this, I'm looking forward to adding this as soon as I can to some of my stuff.

mboudreau:
@iamkirkbater Appreciate the thought and thanks for your opinion.  However, I'm not sure if an array would work if someone would like to create secure values for those environment variables, unless that is possible with arrays, but I'm missing it? Would it be as simple as doing:

```
environment_variables:
  - SOME_VAR:
      secure: "some-secure-string"
```

iamkirkbater:
It actually ends up looking like this:

```environment_variables:
- secure: "some-secure-string"```

where the unencrypted version of `some-secure-string` is something like "KEY=value"

Like if you want to pass through encrypted variables like an AWS_SECRET_ACCESS_KEY to a travis deploy script, you would run `travis encrypt AWS_SECRET_ACCESS_KEY=mysupersecretaccesskey` and then add that value into your .travis.yml file.

https://docs.travis-ci.com/user/environment-variables/#Defining-encrypted-variables-in-.travis.yml

**Comment by mboudreau** [comment]
@iamkirkbater Yeah, I really loathe that "convention", since then you never know which secure string is which (ie, you just want to update a particular var, but don't know which one it is).  I'd much rather have the key human readable, while the value is secured.

**Comment by iamkirkbater** [comment]
And I totally agree with you.

But I just put a comment on the line above saying what the var is. ;)

Anyway, good luck with whatever route you choose :)

**Comment by mboudreau** [comment]
@iamkirkbater I'm not sure I follow.  The key id itself is only known at the time of encryption, not after the fact.  If my coworker needs to update the value of a particular key, they won't be able to without re-generating all values again from scratch since the key id itself is part of the encrypted string.

**Comment by iamkirkbater** [comment]
An example .travis.yml file of mine:

```
env:
  global: 
  - somekey: somevalue
  - anotherKey: anothervalue
  # AWS_ACCESS_KEY_ID
  - secure: "a;dskljfaskdljfa=="
  # AWS_SECRET_ACCESS_KEY
  - secure: "a;lkdfasdjflkasjdf;lkasg="
```

It's an extra step, sure, but it works.

**Comment by mboudreau** [comment]
Eh, I'm weary of comments as a source of truth.  Could easily get screwed up with a merge if not careful.

**Comment by BanzaiMan** [comment]
Sorry for the delay. I'll take a look at this next week.

**Comment by mboudreau** [comment]
@BanzaiMan Anything?

File: lib/dpl/provider.rb [hunk 1]
```diff
@@ -57,6 +57,7 @@ def self.new(context, options)
         name = super.option(:provider).to_s.downcase.gsub(/[^a-z0-9]/, '')
         raise Error, 'could not find provider %p' % options[:provider] unless name = constants.detect { |c| c.to_s.downcase == name }
         provider = const_get(name).new(context, options)
+        puts "Provider #{name} with options #{options}"
```
BanzaiMan: Not sure if we can add this unconditionally. While we do have filter to suppress secrets, some information may be best left unwritten.

lib/dpl/provider/lambda.rb
```diff
@@ -26,23 +26,29 @@ def push_app
...
-              handler:        handler,
-              runtime:        options[:runtime]        || default_runtime,
+              function_name:        function_name,
+              description:          options[:description]    || default_description,
+              timeout:              options[:timeout]        || default_timeout,
+              memory_size:          options[:memory_size]    || default_memory_size,
+              role:                 option(:role),
+              handler:              handler,
+              runtime:              options[:runtime]        || default_runtime,
+              # vpc_config:           vpc_config,
```
BanzaiMan: Are we just adding comments, and aligning the white spaces based on them?

lib/dpl/provider/lambda.rb
```diff
@@ -147,6 +160,31 @@ def output_file_path
...
+
+      def dead_letter_arn
+        options[:dead_letter_arn] ? {target_arn: options[:dead_letter_arn]} : nil
+      end
+
+      def tracing_mode
+        options[:tracing_mode] ? {mode: options[:tracing_mode]} : nil
+      end
+
+      def default_kms_key_arn
```
BanzaiMan: It does not appear that this function and 2 below are invoked anywhere. Did I miss it?

lib/dpl/provider/lambda.rb
```diff
@@ -51,27 +57,34 @@ def push_app
...
-          })
+                                                     function_name:  options[:name] || option(:function_name),
+                                                     zip_file:       function_zip,
+                                                     publish:        publish
+                                                 })
 
           log "Updated code of function: #{response.function_name}."
-        else
+        rescue ::Aws::Lambda::Errors::ResourceNotFoundException
+          log "Function #{function_name} doesn't exists, creating."
```
BanzaiMan: → "does not exist"

spec/provider/lambda_spec.rb
```diff
@@ -89,19 +99,21 @@
...
-        provider.lambda.stub_responses(:list_functions, empty_list_functions_response)
+        provider.lambda.stub_responses(:get_function, 'ResourceNotFoundException')
         provider.lambda.stub_responses(:create_function, example_response)
       end
 
       example do
+
+        expect(provider).to receive(:log).with(/Function #{lambda_options[:function_name]} doesn't exists, creating\./)
         expect(provider).to receive(:log).with(/Created lambda: #{lambda_options[:function_name]}\./)
         provider.push_app
```
BanzaiMan: Looks like some specs are failing here…

lib/dpl/provider.rb
mboudreau: this was mostly used to try to figure out why environment_variables weren't coming through.  It's not showing at this point, which I find perplexing.

lib/dpl/provider/lambda.rb
mboudreau: Again, mostly just trying out things to get the environment variables working.

lib/dpl/provider/lambda.rb
mboudreau: I'll change it.

spec/provider/lambda_spec.rb
mboudreau: Specs are failing because of the extra log info while trying to debug the horrible problems of options being parsed from yaml not working as intended.

mboudreau:
@BanzaiMan I appreciate you looking at the PR, but you completely ignored my original request, which is to figure out *why are environment_variables not being parsed and available through :options?*

mboudreau:
@BanzaiMan Hello? Anyone?

BanzaiMan:
Ooops. Sorry for the delay! I'll take a look at this tomorrow.

BanzaiMan:
The reason that `environment_variables` did not work as it was originally written:

```yaml
deploy:
  provider: lambda
  environment_variables:
    CLIENT_ID:
      secure: "something-secure"
    CLIENT_SECRET:
      secure: "something-secure"
    LOG_LEVEL: "debug"
```

is that each option passed should be a scalar, or an array of scalars. I don't think this is documented anywhere, so we would have to fix that.

If you pass a scalar (e.g., `environment_variables: FOO=BAR`), you'd see this: https://staging.travis-ci.org/BanzaiMan/travis_staging_test/builds/619860#L709. (You can ignore the error.) And similarly for an array of scalars (https://staging.travis-ci.org/BanzaiMan/travis_staging_test/builds/619862#L438)

You _can_ use secure env vars here (e.g., https://github.com/BanzaiMan/travis_staging_test/commit/26b4f181fa3806e073f5e57d527fa7414cb0cdf7); however, we do not filter out the secrets in the logs (because they are not defined at the top level) and it will be shown if you are not careful (https://staging.travis-ci.org/BanzaiMan/travis_staging_test/builds/619864#L709).

Does this help you?

mboudreau:
@BanzaiMan So, what you're saying is that I cannot use a key-value pair without encrypting both key and value.  That really sucks.  If you look up on the comments I had already mentioned that this is not a preferred approach as you then loose the context of what you're trying to encrypt, or move around.

It is at all possible to change this behavior or is there another way around parsing a scalar where the key stays intact while the value is encrypted?

BanzaiMan:
I think you can, if you use the Settings to define the secrets.

In https://staging.travis-ci.org/BanzaiMan/travis_staging_test/builds/620005, I set:

```yaml
deploy:
  provider: lambda
  environment_variables:
    - SECRET=$SUPERDUPER
  ⋮
```

with `SUPERDUPER` set in the settings page (you can see this in https://staging.travis-ci.org/BanzaiMan/travis_staging_test/builds/620005#L417).

The value is passed on to the deployment provider code: https://staging.travis-ci.org/BanzaiMan/travis_staging_test/builds/620005#L590

mboudreau:
@BanzaiMan Is it possible to use the global env in the travis file itself?  I don't like having to separate the key from the actual code and would rather have it in the .travis.yml file.

BanzaiMan:
That should work, but I haven't tested it.

mboudreau:
Okay, thanks for the guidance. Let me try it out and commit my latest once
it works so we can merge this PR

mboudreau:
@BanzaiMan Could you please review my changes again?  Nothing notable from the Lambda side, but I have tested every new config option and it's working marvelously :)

Main thing to note is I've updated the AWS SDK to the latest in V2 since it was missing the `tag_resource` function stub, and I also updated json to 1.8.3 because there are issues with 1.8.1 not able to compile for some reason.

mboudreau:
@BanzaiMan Also, the test works for all my lambda stuff, but there seems to be something wrong with testfairy which I'm not sure when that got introduced.

BanzaiMan:
TestFairy's specs rely on external files at the moment. I think I'll be working on fixing them shortly, so we can ignore those failures for now.

I'll restart other failures with `bundle install`.

BanzaiMan:
Other failures are a bit perplexing. I'll investigate.

mboudreau:
@BanzaiMan Yeah, those other errors are really weird.  It might of just been a hiccup with Travis.  Maybe try restarting the build?  Either that, or checkout my PR and test it locally to make sure it all works.  Like it said, everything was passing locally except for testfairy.

BanzaiMan:
@mboudreau I've decided to skip this brittle spec for now with https://github.com/travis-ci/dpl/commit/b0fed5daad436c59d47ab5b83cfb33afb9d7b658. Please update your PR, so that we can merge and release soon!

mboudreau:
@BanzaiMan Thanks man, it's been updated.

PR #677, opened 2017-07-30
Add option to disable "Deployed at [time]" file

MathiasSM:
Adds a simple option to be set on `.travis.yml` deploy section, just as the others, to be able to disable the creation of such file. It could be deleted with a complicated after_deploy but I don't see why a simple option cannot be provided.

### DIFF
--- a/lib/dpl/provider/pages.rb
+++ b/lib/dpl/provider/pages.rb
@@ -17,6 +17,7 @@ module DPL
         - project-name [optional, defaults to fqdn or repo slug]
         - email [optional, defaults to deploy@travis-ci.org]
         - name [optional, defaults to Deployment Bot]
+        - deployment-file [optional, defaults to false]
       """
 
       require 'tmpdir'
@@ -48,6 +49,8 @@ module DPL
 
         @gh_email = options[:email] || 'deploy@travis-ci.org'
         @gh_name = "#{options[:name] || 'Deployment Bot'} (from Travis CI)"
+
+        @deployment_file = !!options[:deployment_file]
 
         @gh_ref = "#{@gh_url}/#{slug}.git"
         @gh_remote_url = "https://#{@gh_token}@#{@gh_ref}"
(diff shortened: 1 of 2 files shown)


MathiasSM:
I used `"true"` instead of `true` because I thought the yaml got here unparsed, but I'm probably wrong and it should be `true`... Can someone clear that?

sunyang713:
I think this should default to false. Helps keep the `.travis.yml` cleaner. I don't think it makes sense in the first place to make the file, but for those who want it they can enable it.

lib/dpl/provider/pages.rb
```diff
@@ -13,6 +13,7 @@ class Pages < Provider
         - project-name [optional, defaults to fqdn or repo slug]
         - email [optional, defaults to deploy@travis-ci.org]
         - name [optional, defaults to Deployment Bot]
+        - deployment-file [optional, defaults to true]
```
sunyang713: I think this should default to false. Helps keep the .travis.yml cleaner. I don't think it makes sense in the first place to make the file, but for those who want it they can enable it.

polyzen:
Fixes #696

jyasskin:
I'd also just avoid adding the file, but @webknjaz seems to want an option. Is there a use case for having the file that reading the commit message doesn't satisfy?

jyasskin:
If @BanzaiMan wants the change that just removes the file without adding a flag, that's #697.

MathiasSM:
PR now defaults to false (Just in case of a merge?)

lib/dpl/provider/pages.rb
```diff
@@ -33,6 +34,8 @@ def initialize(context, options)
         @gh_email = options[:email] || 'deploy@travis-ci.org'
         @gh_name = "#{options[:name] || 'Deployment Bot'} (from Travis CI)"
 
+        @deployment_file = options[:deployment_file] && (options[:deployment_file]!=false)
```
webknjaz: This is ugly. Try double negation, like `@deployment_file = !!options[:deployment_file]`

BanzaiMan:
Needs conflict resolution.

MathiasSM:
Just incorporated master changes (refactoring of the pages provider script, did not change the way this option works)

polyzen:
It would be nice if this was rebased prior to merge.

lib/dpl/provider/pages.rb
```diff
@@ -17,6 +17,7 @@ class Pages < Provider
         - project-name [optional, defaults to fqdn or repo slug]
         - email [optional, defaults to deploy@travis-ci.org]
         - name [optional, defaults to Deployment Bot]
+        - deployment-file [optional, defaults to false]
```
webknjaz: Please also add this to README

MathiasSM: Optional, defaults to false, enables the creation of deployment-info file?

BanzaiMan:
@MathiasSM Also, https://travis-ci.org/travis-ci/docs-travis-ci-com needs a corresponding PR. Thank you!!

MathiasSM:
I'll see what I can do :)

MathiasSM:
See travis-ci/docs-travis-ci-com#1679 :)

PR #689, opened 2017-09-05
added support for force-updating eb env

kzidane:
Might this small change be of interest? Right now build fails if we push too frequently while an app is still being deployed. This PR allows for force updating of the environment by aborting current environment update (if any), waiting for environment to be ready, then updating the environment. Can be used by adding `force_update: true` to `.travis.yml`.

CC @acnagy


lib/dpl/provider/elastic_beanstalk.rb
```diff
@@ -66,6 +66,10 @@ def env_name
         options[:env] || context.env['ELASTIC_BEANSTALK_ENV'] || raise(Error, "missing env")
       end
 
+      def force_update
```
BanzaiMan: This being boolean, I think `force_update?` is a better name.

lib/dpl/provider/elastic_beanstalk.rb
```diff
@@ -183,11 +187,31 @@ def wait_until_deployed
...
       end
 
+      def abort_update()
+        options = {
+          :environment_name  => env_name
+        }
+
+        begin
+          eb.abort_environment_update(options)
+        rescue; end
```
BanzaiMan: This formatting is weird. Use post-fix `rescue`, or put `end` on a new line.

lib/dpl/provider/elastic_beanstalk.rb
```diff
@@ -183,11 +187,31 @@ def wait_until_deployed
...
+
+        options[:attribute_names] = ["Status"]
+        while eb.describe_environment_health(options).status != "Ready"
+          sleep 5
```
BanzaiMan: Is it possible that this call never returns `Ready`, and this method never returns?

BanzaiMan:
Also, README needs an update.

PR #690, opened 2017-09-17
Relax chef gem constraint for chef-supermarket provider

kamaradclimber:
dpl is now running with ruby 2.2.7 so we don't have to restrain
chef version too much

Also fix compatibility with recent chef version.

### DIFF
--- a/lib/dpl/provider/chef_supermarket.rb
+++ b/lib/dpl/provider/chef_supermarket.rb
@@ -5,10 +5,16 @@ module DPL
       # Most of the code is inspired by:
       # https://github.com/opscode/chef/blob/11.16.4/lib/chef/knife/cookbook_site_share.rb
 
-      # Compatibility with ruby 1.9
-      requires 'rack', version: '< 2.0'
-      requires 'mime-types', version: '~> 1.16'
-      requires 'chef', version: '< 12.0'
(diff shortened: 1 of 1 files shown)


BanzaiMan:
This is not strictly true. While much smaller in number, there are Precise images that still require 1.9 support.

kamaradclimber:
Thanks for the feedback.
Can those images be used for builds targetting `language: ruby` ? (which is used by builds using chef_supermarket provider)

kamaradclimber:
@BanzaiMan would you have any feedback/advice to get this PR merged?

kamaradclimber:
@BanzaiMan what can we do with this PR?

BanzaiMan:
Please test this PR, on both Trusty and Precise.

We are also looking at dropping Ruby 1.9 support from `dpl`, but that may take a little while longer.

kamaradclimber:
I've made a patch which is backward compatible with old ruby versions.

BanzaiMan:
Thanks. Has it been tested?

Annih:
@BanzaiMan I personally tested it on some chef cookbooks. It works fine for me.

BanzaiMan:
With #757, chef_supermarket now resolves to 13.8.0 (as of this writing).

PR #695, opened 2017-09-29
The Pages provider now accepts an encrypted github deploy key.

### LINKED ISSUE #694: Deploy to Github Pages with a deploy key instead of a token
https://github.com/travis-ci/dpl#github-pages currently needs a GitHub oauth token with the repo permission in order to commit to the gh-pages branch. Storing a github token as a TravisCI secret has the side-effect that anyone with write access to the repository can steal the ability to commit to any repository as me. So it'd be nice if the github-pages deployment would let me use a single-repository [deploy key](https://developer.github.com/v3/guides/managing-deploy-keys/#deploy-keys) instead.

We're doing this manually in repositories like https://github.com/heycam/webidl/blob/master/deploy.sh and https://github.com/w3c/permissions/blob/master/deploy.sh. The basic strategy is to use `travis encrypt-file` on the key, and pass the decrypted version to `ssh-add` before trying to `git push`.

I should be able to send you a PR for this if it's something you're willing to take.

jyasskin:
Fixes #694.

Tested in https://travis-ci.org/jyasskin/testing/builds/290726967.

### DIFF
--- a/README.md
+++ b/README.md
@@ -517,6 +517,14 @@ You first need to create an [Atlas account](https://atlas.hashicorp.com/account/
 #### Options:
 
 * **github-token**: GitHub oauth token with `repo` permission.
+* **deploy-key**: Filename of an encrypted
+  [GitHub deploy key](https://developer.github.com/v3/guides/managing-deploy-keys/#deploy-keys)
+  with write access to the **repo** repository, encrypted with `travis
+  encrypt-file`.
+* **deploy-key-key**: The key produced by `travis encrypt-file`, usually of the
+  form `$encrypted_<hex-digits>_key`.
+* **deploy-key-iv**: The IV produced by `travis encrypt-file`, usually of the
+  form `$encrypted_<hex-digits>_iv`.
 * **repo**: Repo slug, defaults to current one.
 * **target-branch**: Branch to push force to, defaults to gh-pages.
 * **keep-history**: Optional, create incremental commit instead of doing push force, defaults to false.
--- a/README.md
+++ b/README.md
@@ -518,13 +518,16 @@ You first need to create an [Atlas account](https://atlas.hashicorp.com/account/
 
 * **github-token**: GitHub oauth token with `repo` permission.
 * **deploy-key**: Filename of an encrypted
-  [GitHub deploy key](https://developer.github.com/v3/guides/managing-deploy-keys/#deploy-keys)
-  with write access to the **repo** repository, encrypted with `travis
-  encrypt-file`.
-* **deploy-key-key**: The key produced by `travis encrypt-file`, usually of the
-  form `$encrypted_<hex-digits>_key`.
-* **deploy-key-iv**: The IV produced by `travis encrypt-file`, usually of the
-  form `$encrypted_<hex-digits>_iv`.
+  [GitHub deploy key](https://developer.github.com/v3/guides/managing-deploy-keys/#deploy-keys) with
+  write access to the **repo** repository, encrypted using
+  [`openssl aes-256-cbc -k`](https://docs.travis-ci.com/user/encrypting-files/#Using-OpenSSL). For
+  example, if your deploy key is in `deploy_key`, its encryption key is in `$DEPLOY_KEY_KEY`, and
+  you want to encrypt it to `deploy_key.enc`, you can use: `openssl aes-256-cbc -k "$DEPLOY_KEY_KEY"
+  -in deploy_key -out deploy_key.enc`.
+* **deploy-key-key**: The decryption key for the deploy-key file. This can be created with `openssl
+  rand -base64 32` and saved securely into Travis CI using `travis encrypt DEPLOY_KEY_KEY=<the
+  key> --add`. Be sure to pick an [encrypted variable](https://docs.travis-ci.com/user/encryption-keys/)
+  name that you're not yet using.
 * **repo**: Repo slug, defaults to current one.
 * **target-branch**: Branch to push force to, defaults to gh-pages.
 * **keep-history**: Optional, create incremental commit instead of doing push force, defaults to false.
@@ -541,6 +544,10 @@ You first need to create an [Atlas account](https://atlas.hashicorp.com/account/
 #### Examples:
 
     dpl --provider=pages --github-token=<api-key> --local-dir=build
+    DEPLOY_KEY_KEY=$(openssl rand -base64 32)
+    travis encrypt DEPLOY_KEY_KEY=$DEPLOY_KEY_KEY --add
+    openssl aes-256-cbc -k $DEPLOY_KEY_KEY -in deploy_key -out deploy_key.enc
+    dpl --provider=pages --deploy-key=deploy_key.enc --deploy-key-key=$DEPLOY_KEY_KEY --local-dir=build
 
 ### GitHub Releases:
 
--- a/README.md
+++ b/README.md
@@ -517,17 +517,9 @@ You first need to create an [Atlas account](https://atlas.hashicorp.com/account/
 #### Options:
 
 * **github-token**: GitHub oauth token with `repo` permission.
-* **deploy-key**: Filename of an encrypted
-  [GitHub deploy key](https://developer.github.com/v3/guides/managing-deploy-keys/#deploy-keys) with
-  write access to the **repo** repository, encrypted using
-  [`openssl aes-256-cbc -k`](https://docs.travis-ci.com/user/encrypting-files/#Using-OpenSSL). For
-  example, if your deploy key is in `deploy_key`, its encryption key is in `$DEPLOY_KEY_KEY`, and
-  you want to encrypt it to `deploy_key.enc`, you can use: `openssl aes-256-cbc -k "$DEPLOY_KEY_KEY"
-  -in deploy_key -out deploy_key.enc`.
-* **deploy-key-key**: The decryption key for the deploy-key file. This can be created with `openssl
-  rand -base64 32` and saved securely into Travis CI using `travis encrypt DEPLOY_KEY_KEY=<the
-  key> --add`. Be sure to pick an [encrypted variable](https://docs.travis-ci.com/user/encryption-keys/)
-  name that you're not yet using.
+* **deploy-key**: A base64-encoded [GitHub deploy
+  key](https://developer.github.com/v3/guides/managing-deploy-keys/#deploy-keys)
+  with write access to the **repo** repository.
 * **repo**: Repo slug, defaults to current one.
 * **target-branch**: Branch to push force to, defaults to gh-pages.
 * **keep-history**: Optional, create incremental commit instead of doing push force, defaults to false.
@@ -544,10 +536,8 @@ You first need to create an [Atlas account](https://atlas.hashicorp.com/account/
 #### Examples:
 
     dpl --provider=pages --github-token=<api-key> --local-dir=build
-    DEPLOY_KEY_KEY=$(openssl rand -base64 32)
-    travis encrypt DEPLOY_KEY_KEY=$DEPLOY_KEY_KEY --add
-    openssl aes-256-cbc -k $DEPLOY_KEY_KEY -in deploy_key -out deploy_key.enc
-    dpl --provider=pages --deploy-key=deploy_key.enc --deploy-key-key=$DEPLOY_KEY_KEY --local-dir=build
+    travis encrypt DEPLOY_KEY="$(cat deploy_key|base64)"
+    dpl --provider=pages --deploy-key="$DEPLOY_KEY" --local-dir=build
 
 ### GitHub Releases:
 
--- a/README.md
+++ b/README.md
@@ -517,7 +517,7 @@ You first need to create an [Atlas account](https://atlas.hashicorp.com/account/
 #### Options:
 
 * **github-token**: GitHub oauth token with `repo` permission.
-* **deploy-key**: A base64-encoded [GitHub deploy
+* **deploy-key**: A base64-encoded [GitHub deploy private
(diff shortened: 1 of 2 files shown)


lib/dpl/provider/pages.rb
```diff
@@ -60,17 +74,31 @@ def github_deploy
...
         context.shell 'git add .'
         context.shell "git commit -m 'Deploy #{@project_name} to #{@gh_ref}:#{@target_branch}'"
-        context.shell "git push --force --quiet 'https://#{@gh_token}@#{@gh_ref}' master:#{@target_branch} > /dev/null 2>&1"
+        if @gh_token
+          target_repo = "https://#{@gh_token}@#{@gh_ref}"
+        else
+          # Depend on the key set up in push_app.
+          target_repo = "git@github.com:#{slug}.git"
+        end
+        context.shell "git push --force --quiet '#{target_repo}' master:#{@target_branch}"
```
jyasskin: I don't absolutely need to remove the `> /dev/null 2>&1` for this change, but it makes errors much easier to debug.

BanzaiMan:
Do you have a test build for this PR?

jyasskin:
https://travis-ci.org/jyasskin/testing/builds/281468169

README.md
```diff
@@ -517,6 +517,14 @@ You first need to create an [Atlas account](https://atlas.hashicorp.com/account/
 #### Options:
 
 * **github-token**: GitHub oauth token with `repo` permission.
+* **deploy-key**: Filename of an encrypted
+  [GitHub deploy key](https://developer.github.com/v3/guides/managing-deploy-keys/#deploy-keys)
+  with write access to the **repo** repository, encrypted with `travis
+  encrypt-file`.
+* **deploy-key-key**: The key produced by `travis encrypt-file`, usually of the
```
BanzaiMan: Why is this necessary?

This discourages use with the repositories which already has encrypted file. https://docs.travis-ci.com/user/encrypting-files#Encrypting-multiple-files It is cumbersome, if not impossible, to add this feature.

jyasskin: Oh geez, that's awful. Why wouldn't it just add a new $encrypted_* pair? 

Anyway, it's clearly necessary to encrypt the deploy key: otherwise anyone with read access to the repository could also write to it. I can see a couple ways to avoid travis-ci/travis.rb#239:

1. Point people to https://docs.travis-ci.com/user/encrypting-files#Using-OpenSSL to generate the encrypted deploy-key file, and provide two parameters, one pointing at the file, and the other to the decryption key. I like this option best, so I'll start on it.
1. Have folks decrypt the deploy key in their `before_deploy` steps, and just point to it here. This has the two downsides that 1) it's more code to write in `.travis.yml`, and 2) it makes it easy to use an insecure deploy key checked directly into the repository.
1. Have two variables to find the deploy key file: 1 identifying the encrypted tarfile, and the other pointing to the deploy key inside it.

jyasskin: Done. New test build at https://travis-ci.org/jyasskin/testing/builds/290726967.

BanzaiMan: It is still not clear on *why* the use of "deploy key" as a separate file is necessary. In other deployment providers, deployment keys are given as a bare string (whether or not it is encrypted). For example,

```yaml
deploy:
  provider: releases
  api_key:
    secure: YOUR_API_KEY_ENCRYPTED
⋮
```

It is not clear to me why this provider *requires* an encrypted file to deploy (or have that option at all). If the requirement is *having* the `#{gitdir}/deploykey` file, then it is easy to write that file from what is given in the configuration.

lib/dpl/provider/pages.rb
webknjaz: @jyasskin I've added `--debug` support (AFAIR natively supported by dpl) to my feature branch for this :)

jyasskin:
Any word on this change?

webknjaz:
I think we should also ref https://github.com/travis-ci/dpl/issues/694#issuecomment-347092770, so it won't get lost.

@BanzaiMan is there a way to locate/reuse that autogenerated private key?

lib/dpl/provider/pages.rb
```diff
@@ -45,9 +49,15 @@ def slug
       end
 
       def check_auth
+        unless @gh_token || @gh_deploy_key
```
webknjaz: I've implemented this method for token version by talking to the api in #719.
You'll need to extend it with smth like `ssh -Tvvvvv git@github.com` to check whether the key is valid :)

lib/dpl/provider/pages.rb
```diff
@@ -45,9 +49,15 @@ def slug
...
+          error "must specify github-token or deploy-key"
+        end
       end
 
       def needs_key?
+        # Provider's generic support for keys is about provisioning and
```
webknjaz: Is this documented somewhere? This should probably return `true` if there's no `gh-token` provided.

lib/dpl/provider/pages.rb
```diff
@@ -60,17 +70,32 @@ def github_deploy
         context.shell "echo '#{@gh_fqdn}' > CNAME" if @gh_fqdn
         context.shell 'git add .'
         context.shell "FILES=\"`git commit -m 'Deploy #{@project_name} to #{@gh_ref}:#{@target_branch}' | tail`\"; echo \"$FILES\"; echo \"$FILES\" | [ `wc -l` -lt 10 ] || echo '...'"
-        context.shell "git push --force --quiet 'https://#{@gh_token}@#{@gh_ref}' master:#{@target_branch} > /dev/null 2>&1"
+        if @gh_token
+          target_repo = "https://#{@gh_token}@#{@gh_ref}"
+        else
+          # Depend on the key set up in push_app.
+          target_repo = "git@github.com:#{slug}.git"
```
webknjaz: It's not always `github.com` and `@gh_ref = "#{@gh_url}/#{slug}.git"` (see above). So you could probably do smth like:

```ruby
url_start = 'git' if ssh_key is given else "https://#{@gh_token}"
url_delimiter = ':' if ssh_key is given else '/'
target_repo = "#{url_start}@#{@gh_url}#{url_delimiter}#{slug}.git"
```

But I'd store `@gh_ref` in init.

remyabel:
Sorry, I'm a bit out of the loop. Why are the deploy keys base64-encoded? Wouldn't it make more sense to take input from a file? And if they are public keys, is there any reason that they need to be kept private? I'm under the impression that the private keys (on the Travis CI side) is what needs to be encrypted/kept private.

webknjaz:
@remyabel I think, your comment is quite useful and shows that option naming can be misleading :)

Anyway, I'd like to avoid requiring user to generate/add key at all. See https://github.com/travis-ci/dpl/issues/694#issuecomment-347092770.

remyabel:
Yeah the more I read it the more I get confused. The use of "deploy key" and linking to the Github help suggests it's a private key, while `setup_git_ssh` uses the `-i` switch which definitely reads in a private key. I would personally rename it, but the projects referenced in this PR call it "deploy key", which might be common usage for it.

jyasskin:
@webknjaz @remyabel The [github docs](https://developer.github.com/v3/guides/managing-deploy-keys/#deploy-keys) consistently use "deploy key" to refer to the private half of the keypair and "public key" to refer to the public half. For example, knowledge of a public key can't be used to "grant access" to anything unless you want the public to have access, and there's no reason to mention that a public key is "not protected by a passphrase" since they never are.

I do see how it could be confusing that the actual Github settings page just says "Deploy keys / Add new" and "Key" instead of "public key", and I can be more clear that --deploy-key takes the private key in the documentation.

remyabel:
Roger, that makes sense.

webknjaz:
@jyasskin my PR has been merged. Plz make sure to rebase your changes :)

lib/dpl/provider/pages.rb
```diff
@@ -53,7 +58,12 @@ def initialize(context, options)
         @deployment_file = !!options[:deployment_file]
 
         @gh_ref = "#{@gh_url}/#{slug}.git"
-        @gh_remote_url = "https://#{@gh_token}@#{@gh_ref}"
+        if @gh_token
+          @gh_remote_url = "https://#{@gh_token}@#{@gh_ref}"
+        else
+          # Depend on the key set up in push_app.
+          @gh_remote_url = "git@#{@gh_ref}:#{slug}.git"
```
webknjaz: gh_ref -> gh_url

lib/dpl/provider/pages.rb
```diff
@@ -109,25 +119,55 @@ def setup_auth
...
-        error "gh-token is invalid. Details: #{exc}"
+            log "Logged in as @#{user.login} (#{user.name})"
+          rescue Octokit::Unauthorized => exc
+            error "gh-token is invalid. Details: #{exc}"
+          end
+        elsif  @gh_deploy_key
+          Dir.mktmpdir do |tmpdir|
+            github_install_key(tmpdir)
+            context.shell "ssh -T git@#{@gh_url}"
+            if $?.exitstatus == 255
```
webknjaz: I'd say: != 0

lib/dpl/provider/pages.rb
```diff
@@ -109,25 +119,55 @@ def setup_auth
...
+            end
 
-        log "Logged in as @#{user.login} (#{user.name})"
-      rescue Octokit::Unauthorized => exc
-        error "gh-token is invalid. Details: #{exc}"
+            log "Logged in as @#{user.login} (#{user.name})"
+          rescue Octokit::Unauthorized => exc
+            error "gh-token is invalid. Details: #{exc}"
+          end
+        elsif  @gh_deploy_key
```
webknjaz: Nit: two spaces

lib/dpl/provider/pages.rb
webknjaz: This is still unresolved

lib/dpl/provider/pages.rb
```diff
@@ -109,25 +119,55 @@ def setup_auth
...
+          Dir.mktmpdir do |tmpdir|
+            github_install_key(tmpdir)
+            context.shell "ssh -T git@#{@gh_url}"
+            if $?.exitstatus == 255
+              error "deploy-key is invalid"
+            end
+          end
+        else
+          error "must specify github-token or deploy-key"
```
webknjaz: I guess this should go to `needs_key`

lib/dpl/provider/pages.rb
```diff
@@ -109,25 +119,55 @@ def setup_auth
...
+        # provided by the caller.
         false
       end
 
       def print_step(msg)
         log msg if @verbose
       end
 
+      def github_install_key(tmpdir)
+        if @gh_deploy_key
```
webknjaz: You already do this check when calling this method. I think it's not good to silence errors this way

lib/dpl/provider/pages.rb
```diff
@@ -188,6 +228,8 @@ def push_app
         print_step "Starting deployment of #{@target_branch} branch to GitHub Pages..."
         print_step "The deployment is configured to preserve the target branch if it exists on remote" if @keep_history
         Dir.mktmpdir do |tmpdir|
+            github_install_key(tmpdir)
```
webknjaz: This probably should go to `needs_key` and be called only once, instead of doing this in two places.

lib/dpl/provider/pages.rb
```diff
@@ -53,7 +58,12 @@ def initialize(context, options)
         @deployment_file = !!options[:deployment_file]
 
         @gh_ref = "#{@gh_url}/#{slug}.git"
-        @gh_remote_url = "https://#{@gh_token}@#{@gh_ref}"
+        if @gh_token
```
webknjaz: Honestly, I would swap it at made a deployment key take precedence over token.

lib/dpl/provider/pages.rb
```diff
@@ -109,25 +119,52 @@ def setup_auth
...
+        Dir.mkdir(gitdir)
+        print_step "Decoding git deploy key into #{gitdir}"
+        # Install the key.
+        File.open("#{gitdir}/deploykey", "w", 0600) {|file|
+          file.write Base64.decode64(@gh_deploy_key)
+        }
+        setup_git_ssh("#{gitdir}/git-ssh", "#{gitdir}/deploykey")
+        # Return true if the deploy key actually works.
+        context.shell "ssh -i .dpl/git/deploykey -T git@#{@gh_url}"
+        return $?.exitstatus != 255
```
webknjaz: Alright, I would probably check output for the welcome message to be 100% sure.

BanzaiMan:
Hello, there. Sorry for the delay. I'm reorganizing code for the 1.9.0 release, so PRs are on hold until I can get that out of the way.

BanzaiMan:
1.9.0 is out. If you are still interested, please either update this PR, or open a new one.

Thank you!

jyasskin:
It's going to take me a couple weeks to rebase (busy with the IETF meeting), but I'm definitely still interested.

webknjaz:
@jyasskin hey, how's it going?

webknjaz:
@jyasskin?

svenfuchs:
Hey @jyasskin :wave:

I have ported your work on this branch to a new PR here: https://github.com/travis-ci/dpl/pull/1031. I believe this should be equivalent to your work.

If you would like me to amend the commit in order to add your Git user name and email address please let me know. (My commit was based on your work, so I would happily pass authorship for it, but I also did not want to presume.)

It would be great if you could help us test if the issue that your change was meant to address. You can opt into using dpl v2 by adding this to your `.travis.yml` file:

```yaml
deploy:
  provider: [provider-name]
  edge:
    branch: master
⋮
```

Thank you!

webknjaz:
@svenfuchs I think this can be closed now.

svenfuchs:
right. thanks @webknjaz!

Issue #696, opened 2017-09-30
Remove undesired and forced file creation in GitHub Pages deployment


sunyang713:
The GitHub Pages deployment is polluted with an empty file with the title of the deployment, i.e.: "deployed at Mon Sep 18 02:58:23 UTC 2017 by Deployment Bot (from Travis CI)"

It doesn't make sense to create a file with info about the build that is exposed publicly with the rest of the deployment, and it feels wrong that the deployment process would force the inclusion of such an unnecessary file. At the very least, it should be an opt-in feature, opted-out by default.

[Here is the statement](https://github.com/travis-ci/dpl/blob/master/lib/dpl/provider/pages.rb#L56)

The commit message associated with every GitHub Pages deployment is detailed and sufficient for expressing the Travis deployment.

BanzaiMan:
https://github.com/travis-ci/dpl/pull/677 adds an option for this.

PR #697, opened 2017-09-30
Don't create empty file in GitHub Pages Deployment

sunyang713:
Addresses #696


jyasskin:
This is a duplicate of #677.

sunyang713:
ah, thanks. closing then.

PR #713, opened 2017-11-11
Drop verification of existing firebase-tools installation

sylveon:
This prevents https://github.com/travis-ci/travis-ci/issues/8637 from happening.

The check uses the PATH, which in case of node.js projects includes `node_modules/.bin`. If the firebase-tools package has been installed locally, the check will pass because they are stored in said folder, so the tools will not get installed globally. The problem is that when the deploy is run, `node_modules/.bin` isn't in the PATH so it results in 'sh: 1: firebase: not found' when deploying.

### DIFF
--- a/lib/dpl/provider/firebase.rb
+++ b/lib/dpl/provider/firebase.rb
@@ -1,7 +1,7 @@
 module DPL
   class Provider
     class Firebase < Provider
-      npm_g 'firebase-tools@^3.0', 'firebase'
+      npm_g 'firebase-tools@^3.0'
 
       def check_auth
         raise Error, "must supply token option or FIREBASE_TOKEN environment variable" if !options[:token] && !context.env['FIREBASE_TOKEN']


BanzaiMan:
I have a feeling that this is not the right fix. If the mismatch between RVM's `PATH` and <tt>Kernel#`</tt>'s `PATH` is a general problem (not just Firebase), it should be fixed in `DPL::Provider`.

sylveon:
I mean, that'd be ideal, sure. But then either:
- We add `node_modules/.bin` to the PATH everywhere, issues like https://github.com/travis-ci/travis-ci/issues/5092 might become more and more common.
- We remove `node_modules/.bin` from the PATH everywhere, and we break a lot of CI builds.
- We put `node_modules/.bin` at the end of PATH instead of the beginning and make this PATH change everywhere. This will probably break a few CI builds, but not as much as solution 2.

BanzaiMan:
The version specified in this PR is too old at this point.

svenfuchs:
Hey @sylveon :wave:

I believe this PR here addresses your issue by adding `node_modules/.bin` to the `PATH` https://github.com/travis-ci/dpl/pull/1033. I believe this should be an adequate fix, what do you think?

Thank you!

sylveon:
It should, but I don't think you linked the right PR, as it's completely unrelated to node.

svenfuchs:
@sylveon lol, sorry about that. juggling too many PRs at the moment. here's the right one: https://github.com/travis-ci/dpl/pull/1032

PR #714, opened 2017-11-13
Silence curl output when not running from TTYs

### LINKED ISSUE #576: Silence curl output for heroku deployments
It would be very nice to provide an option to silence curl output when uploading application to avoid all this noise in CI logs

image

Maybe default value can be based on `$stdout.isatty` value, in this way typical user does not need to configure anything.

fabn:
Closes #576


BanzaiMan:
Thanks for the PR. Has this been tested?

BanzaiMan:
Closing this in favor of #888.

PR #715, opened 2017-11-19
Preserve history when deploying to gh-pages.

### LINKED ISSUE #704: Allow github pages deployments to retain history
The [W3C deployment scripts](https://gist.github.com/domenic/ec8b0fc8ab45f39403dd) that I'm trying to replace retain the history of the gh-pages branch.

If this is a change you'd accept, would you rather an option to retain history, or just unconditionally do it?

jyasskin:
Fixes #704.

This change sits on top of #695, which is mostly reviewed.

@webknjaz, does this look reasonable to you?

Test deploys:
* https://travis-ci.org/jyasskin/testing/builds/304151299 testing a deploy with no existing gh-pages branch.
* https://travis-ci.org/jyasskin/testing/builds/304151521 testing a deploy with an existing gh-pages branch
* https://travis-ci.org/jyasskin/testing/builds/304151699 testing that a no-op deploy is skipped.

You can see the resulting gh-pages branch at https://github.com/jyasskin/testing/commits/gh-pages.

### DIFF
--- a/lib/dpl/provider/pages.rb
+++ b/lib/dpl/provider/pages.rb
@@ -84,9 +84,12 @@ module DPL
       end
 
       def github_deploy
-        context.shell "touch \"deployed at `date` by #{@gh_name}\""
         context.shell "echo '#{@gh_fqdn}' > CNAME" if @gh_fqdn
         context.shell 'git add .'
+        if context.shell "git diff --cached --quiet"
+          log "No changes; skipping deploy."
+          return true
+        end
         context.shell "FILES=\"`git commit -m 'Deploy #{@project_name} to #{@gh_ref}:#{@target_branch}' | tail`\"; echo \"$FILES\"; echo \"$FILES\" | [ `wc -l` -lt 10 ] || echo '...'"
         # The following line will fail if another deploy happened concurrently.
         # We'd rather have the deploy for the later revision succeed, but that
--- a/lib/dpl/provider/pages.rb
+++ b/lib/dpl/provider/pages.rb
@@ -1,3 +1,5 @@
+require "base64"
+
 module DPL
   class Provider
     class Pages < Provider
@@ -6,8 +8,7 @@ module DPL
       Options:
         - repo [optional, for pushed to other repos]
         - github-token [this or deploy-key required]
-        - deploy-key [this or github-token required; deploy key encrypted with `openssl aes-256-cbc -k`]
-        - deploy-key-key [required if deploy-key is present]
+        - deploy-key [this or github-token required]
         - github-url [optional, defaults to github.com]
         - target-branch [optional, defaults to gh-pages]
         - local-dir [optional, defaults to `pwd`]
@@ -35,9 +36,6 @@ module DPL
         unless @gh_token || @gh_deploy_key
           raise(Error, "must specify github-token or deploy-key")
         end
-        if @gh_deploy_key
-          @gh_deploy_key_key = option(:deploy_key_key)
-        end
 
         @gh_email = options[:email] || 'deploy@travis-ci.org'
         @gh_name = "#{options[:name] || 'Deployment Bot'} (from Travis CI)"
@@ -58,8 +56,8 @@ module DPL
 
       def needs_key?
         # Provider's generic support for keys is about provisioning and
-        # deprovisioning them, while this provider needs to install an encrypted
-        # key provided by the caller.
+        # deprovisioning them, while this provider needs to install a key
+        # provided by the caller.
         false
       end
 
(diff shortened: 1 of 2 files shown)


BanzaiMan:
Thanks for the PR. Could you explain why you are proposing `deploy-key` and `deploy-key-key`? These names don't feel particularly self-revealing, and forcing encryption seems rather strange.

jyasskin:
@BanzaiMan I explained the need for encryption in https://github.com/travis-ci/dpl/pull/695#discussion_r146095804 item 2.

BanzaiMan:
Oh, sorry. If this is dependent on #695, could you add to that PR instead of opening a new one? Thanks.

jyasskin:
They're totally different features; they just have merge conflicts. If you could review #695 with less than a month latency, this wouldn't need to sit on it.

lib/dpl/provider/pages.rb
```diff
@@ -28,7 +30,14 @@ def initialize(context, options)
 
         @gh_fqdn = fqdn
         @gh_url = options[:github_url] || 'github.com'
-        @gh_token = option(:github_token)
+        @gh_token = options[:github_token]
+        @gh_deploy_key = options[:deploy_key]
+        unless @gh_token || @gh_deploy_key
```
webknjaz: This must go to the `check_auth` method

lib/dpl/provider/pages.rb
```diff
@@ -48,29 +57,70 @@ def check_auth
...
+            setup_git_ssh("#{gitdir}/git-ssh", "#{gitdir}/deploykey")
+          end
+          Dir.mktmpdir {|tmpdir|
+            # A failed `git clone` removes its target directory, so use a
+            # subdirectory to make sure we can safely re-create it.
+            tmpdir = "#{tmpdir}/deploy"
+            unless create_git_dir(tmpdir)
+              error "Couldn't create checkout to push to #{@gh_ref}:#{@target_branch}"
+            end
+            context.shell "rsync -r --exclude .git --delete #{@build_dir}/ #{tmpdir}" or error "Could not copy #{@build_dir}."
```
webknjaz: why rsync?

webknjaz:
@jyasskin sorry, I was offline for a while. I'll try reviewing this and #695 in detail soon. Did you backport some commits from https://github.com/webknjaz/dpl/tree/feature/enable-gh-pages-keep-history?

lib/dpl/provider/pages.rb
```diff
@@ -48,29 +57,70 @@ def check_auth
...
+
       def github_deploy
-        context.shell 'rm -rf .git > /dev/null 2>&1'
-        context.shell "touch \"deployed at `date` by #{@gh_name}\""
-        context.shell 'git init' or raise 'Could not create new git repo'
-        context.shell "git config user.email '#{@gh_email}'"
-        context.shell "git config user.name '#{@gh_name}'"
         context.shell "echo '#{@gh_fqdn}' > CNAME" if @gh_fqdn
         context.shell 'git add .'
+        if context.shell "git diff --cached --quiet"
```
webknjaz: I'd prefer to give user more control over an empty commit https://github.com/webknjaz/dpl/commit/c8201a93163142dc43245ec5c097d6a054307245, having reasonable default, of course.

webknjaz:
Also, I'd prefer this to be merged before #695 and be decoupled.

lib/dpl/provider/pages.rb
```diff
@@ -86,9 +86,12 @@ def create_git_dir(gitdir)
       end
 
       def github_deploy
-        context.shell "touch \"deployed at `date` by #{@gh_name}\""
```
webknjaz: This change is unrelated and must be feature-flagged

webknjaz:
@jyasskin I've created PR #719 to showcase you my old code. Please review it. I feel like it's better in terms of public API

lib/dpl/provider/pages.rb
jyasskin: Done in 7f4184d4f19bb2191106f05cbf260a95b78fece3.

lib/dpl/provider/pages.rb
jyasskin: That's plausible, but new options have a cost, and my impression is that @BanzaiMan doesn't want this one. I'm happy to add this option if they prefer to have it. See also https://github.com/travis-ci/dpl/pull/719#pullrequestreview-79236861.

lib/dpl/provider/pages.rb
jyasskin: It simplifies the process of removing the source .git file and the destination files that aren't present in the source. More discussion in https://github.com/travis-ci/dpl/pull/719#discussion_r153264248 and https://github.com/travis-ci/dpl/pull/719#discussion_r153267193.

153957:
As I mentioned in #719:
> I am in favor if keeping the option to force push (discard history), I do not care which way will be the default, but we have some repositories in which the changes in the gh-pages branch are quite large (binary files) and it is not important to keep the history, so to keep the repo size down discarding history is a great feature.

BanzaiMan:
This can now be closed, since #719 has been merged, right?

webknjaz:
Correct.

PR #716, opened 2017-11-20
PyPI: Drop using deprecated ez_setup.py in favor of get-pip.py.

### LINKED ISSUE #603: pypi deployment uses deprecated "ez_setup.py" installer for setuptools
The recommended way to install setuptools is via pip.
The https://bootstrap.pypa.io/ez_setup.py script is now deprecated, and can only install setuptools==33.1.1, which is the last version that supports self upgrade/installation.

Check https://github.com/pypa/setuptools/issues/581 for more details.

jezdez:
Fix #603.

This also removes installing wheel conditionally since get-pip.py will
install wheel automatically (see https://pip.pypa.io/en/stable/installing/).

More info about the deprecation of ez_setup.py: https://github.com/pypa/setuptools/issues/581

### DIFF
--- a/lib/dpl/provider/pypi.rb
+++ b/lib/dpl/provider/pypi.rb
@@ -32,21 +32,15 @@ module DPL
           (options.has_key?(:skip_upload_docs) && options[:skip_upload_docs])
       end
 
-      def self.install_setuptools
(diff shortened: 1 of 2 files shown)


jezdez:
@dstufft Does that make sense to you?

jezdez:
Closing as this is now in #717.

PR #717, opened 2017-11-20
PyPI: Raise error when twine upload fails and use get-pip.py.

### LINKED ISSUE #190: PyPi deployment doesn't recognize failiures
For example [this](https://travis-ci.org/m1keil/test_travis_deploy/builds/38809760]) deployment.

Upload fails:

``` shell
Upload failed (400): A file named "app-0.2.tar.gz" already exists for  app-0.2. To fix problems with that file you should create a new release.
error: Upload failed (400): A file named "app-0.2.tar.gz" already exists for  app-0.2. To fix problems with that file you should create a new release.
```

I was expecting to get some kind of notification about this.

On command line it looks pretty easy to detect failiure:

``` shell
-> python setup.py sdist upload -r test
running sdist
...
...
Submitting dist/app-0.2.tar.gz to https://testpypi.python.org/pypi
Upload failed (400): A file named "app-0.2.tar.gz" already exists for  app-0.2. To fix problems with that file you should create a new release.
error: Upload failed (400): A file named "app-0.2.tar.gz" already exists for  app-0.2. To fix problems with that file you should create a new release.
test_travis_deploy (pypi-release) -> echo $?
1
```

I tried to catch this error with the `after_deploy` hook, but no success.

jezdez:
Fix #190, #567 and #603.

This also removes installing wheel conditionally since get-pip.py will
install wheel automatically (see https://pip.pypa.io/en/stable/installing/).

More info about the deprecation of ez_setup.py: https://github.com/pypa/setuptools/issues/581


jezdez:
@BanzaiMan This would be splendid to have, I hope the way to catch a exit code of 1 is right.

edmorley:
Hi! I'm presuming this would have helped with this?
https://travis-ci.org/aio-libs/aiohttp/jobs/329243597#L2262
(ref https://github.com/aio-libs/aiohttp/issues/2342#issuecomment-357922139)

webknjaz:
@BanzaiMan could you please accept this PR?

BanzaiMan:
This introduces some spec failures.

BanzaiMan:
Also, see https://github.com/travis-ci/dpl/pull/716#issuecomment-345727780.

lib/dpl/provider/pypi.rb
```diff
@@ -32,23 +32,13 @@ def skip_upload_docs?
...
-      end
-
-      def initialize(*args)
-        super(*args)
-        self.class.pip 'wheel' if pypi_distributions.to_s.include? 'bdist_wheel'
+      def install_deploy_dependencies
+        unless context.shell 'wget https://bootstrap.pypa.io/get-pip.py -O - | sudo python'
+          error "Couldn't install pip and setuptools."
+        end
+        context.shell("pip install twine", retry: true) if `which twine`.chop.empty?
```
webknjaz: Why not upgrade pip and twine to latest versions at this point?

jezdez: `get-pip.py` will already [install the latest versions of setuptools, wheel and pip](https://github.com/pypa/get-pip/blob/430ba37776ae2ad89f794c7a43b90dc23bac334c/template.py#L170-L172) itself and running `pip install twine` effectively means installing the most recent version from PyPI.

edmorley: (comment edited to correct line numbers in source link)

jezdez: Good catch, so I'd suggest to do:
```
get-pip.py --no-setuptools --no-wheel
pip install --upgrade setuptools twine wheel
```

edmorley: I could have misread, but I think this would work too:
`get-pip.py setuptools wheel twine`

edmorley: Though your way is less reliant on get-pip.py quirks so perhaps best.

jezdez: You're right, I was confused by how the `implicit_*` flag where handled and that it would take sys.argv straight to do the call with `--upgrade`. The code around this is slightly weird to make sure that there isn't a double entry of setuptools or wheel in the args.

jezdez:
@BanzaiMan Ok, could you take a look at the changes? This includes #716 as well now.

webknjaz:
@jezdez [this doc](https://github.com/travis-ci/dpl/blob/cdab1c3/TESTING.md#testing-dpl-in-the-context-of-travis-ci-builds) requires that you provide a link to a test build using your branch via `edge` option.

jezdez:
@webknjaz Thanks for the pointer, I didn't know about that keyword.

I created the project [travis-ci-pypi-deploy-test](https://github.com/jezdez/travis-ci-pypi-deploy-test/) and set it up to auto-deploy on tags using the pypi-error branch in my dpl fork.

Here's the successful build and deploy: https://travis-ci.org/jezdez/travis-ci-pypi-deploy-test/builds/330851628
Here's the release on PyPI: https://pypi.python.org/pypi/travis-ci-pypi-deploy-test/0.1

jezdez:
FWIW there was a transient error with the Heroku provider that got fixed by running the build for this PR again.

lib/dpl/provider/pypi.rb
```diff
@@ -32,23 +32,15 @@ def skip_upload_docs?
...
-
-      def self.install_twine
-        shell("pip install twine", retry: true) if `which twine`.chop.empty?
-      end
-
-      def initialize(*args)
-        super(*args)
-        self.class.pip 'wheel' if pypi_distributions.to_s.include? 'bdist_wheel'
+      def install_deploy_dependencies
+        unless context.shell "wget https://bootstrap.pypa.io/get-pip.py && " \
```
webknjaz: @jezdez 
instead of storing python script in the fs you could use pipe:
```shell
wget -O - https://bootstrap.pypa.io/get-pip.py | python - --no-setuptools --no-wheel
```

jezdez: Ok. I think we've optimized this patch enough 😁

webknjaz:
@BanzaiMan this PR is now production-ready

jezdez:
Huzzah! 🙌

lib/dpl/provider/pypi.rb
```diff
@@ -32,23 +32,13 @@ def skip_upload_docs?
...
-      def self.install_twine
-        shell("pip install twine", retry: true) if `which twine`.chop.empty?
-      end
-
-      def initialize(*args)
-        super(*args)
-        self.class.pip 'wheel' if pypi_distributions.to_s.include? 'bdist_wheel'
+      def install_deploy_dependencies
+        unless context.shell "wget -O - https://bootstrap.pypa.io/get-pip.py | python - --no-setuptools --no-wheel && " \
+                             "pip install --upgrade setuptools twine wheel"
```
BanzaiMan: Previously, we were using `sudo` to run `ez_setup.py`. https://github.com/travis-ci/travis-ci/issues/9139 reports that this perhaps leads to "permission denied" error. Is it feasible to do the same here?

webknjaz: Ideally, this should run in the same virtualenv where the tests were running. I believe, this happens to `language: python`. Is it possible to have this in other envs? It is a bad practice to influence system interpreter installation. `dpl` is being run in envs other than Travis. It would make me feel terribly uncomfortable, if such deployment tool would ask me to execute smth as root. I think, this should be at least limited to user dir.

webknjaz: If it's really required to be run as root, why not run the whole thing with sudo? This would be explicit and will be run by the user mindfully.

PR #719, opened 2017-11-26
Enable gh-pages to retain deploy branch history

### LINKED ISSUE #704: Allow github pages deployments to retain history

webknjaz:
Closes travis-ci/travis-ci#7562
Resolves #704 

This duplicates efforts of #715, but a bit differently. I wrote it a long time ago, but was unsure whether it's ready. Now I think it's better to present this code to public at least.

The design of this feature is also based on my analysis of [some npm packages](https://www.npmjs.com/search?q=gh-pages) providing similar behavior.

@jyasskin please check it out.

-------------
##### Test run:
https://travis-ci.org/webknjaz/dpl-gh-pages/builds/323649670
(It's got checks for non-existing branch on remote, for enabled and for disabled force push, for `committer-from-gh` option, and check for invalid token — that's why it's errored in the end and it's okay)

##### Command-line options added (some as minor semi-related improvements):
* **keep-history**: Optional, create incremental commit instead of doing push force, defaults to false.
* **allow-empty-commit**: Optional, defaults to false. Enabled if only keep-history is true.
* **committer-from-gh**: Optional, defaults to false. Allows to use token's owner name and email for commit. Overrides `email` and `name` options.
* **verbose**: Optional, be verbose about internal steps, defaults to false.

##### Decisions made:
- **keep-history** must be switchable, because force pushes may help eliminate repo growth in case of lots of blobs (https://github.com/travis-ci/dpl/pull/719/files#r155931757).
- **allow-empty-commit** must be switchable, because of my personal wish to support different cases I saw in similar solutions/tutorials and I think we shouldn't be breaking possible preferences regarding this as build indicator.
- **committer-from-gh**: I added this, because it became possible after introducing auth_check improvement adding an API client. And I think it might be handy.
- I reused existing **verbose** flag to add switchable logging of what's happening during script run, because logs are the best cure from debugging.
- I've added auth_check to give end-users better clue on what's wrong with their token and as a side-effect I've logging the token user's name/handle.
- I'm maintaining repo copy in subfolder as advised @jyasskin, because some of `git clone` removing destination folder on fail and we don't want to lose ownership over tmpdir (which is controlled by the OS).
- This provider supports cases of pre-existing branch as well of non-existent. It creates an orphan branch if creating it, and it is not switchable. If anyone would need non-orphan it should be easy to extend this behavior.
- I've replaced a dirty hack of trimming of commit command output using just git.
- `git add -A .` will still have explicit `-A`, because different versions may have different defaults, which would lead to flakiness (https://github.com/travis-ci/dpl/pull/719/files#r155342614).

##### Side-effect finding:
- It's better not to use `&>` shell stream redirection syntax when spawning commands from within Ruby script: it's broken (#730).

### DIFF
--- a/README.md
+++ b/README.md
@@ -519,6 +519,8 @@ You first need to create an [Atlas account](https://atlas.hashicorp.com/account/
 * **github-token**: GitHub oauth token with `repo` permission.
 * **repo**: Repo slug, defaults to current one.
 * **target-branch**: Branch to push force to, defaults to gh-pages.
+* **keep-history**: Optional, create incremental commit instead of doing push force, defaults to false.
+* **verbose**: Optional, be verbose about internal steps, defaults to false.
 * **local-dir**: Directory to push to GitHub Pages, defaults to current.
 * **fqdn**: Optional, no default, sets a main domain for your website.
 * **project-name**: Defaults to fqdn or repo slug, used for metadata.
--- a/README.md
+++ b/README.md
@@ -521,12 +521,13 @@ You first need to create an [Atlas account](https://atlas.hashicorp.com/account/
 * **target-branch**: Branch to push force to, defaults to gh-pages.
 * **keep-history**: Optional, create incremental commit instead of doing push force, defaults to false.
 * **allow-empty-commit**: Optional, defaults to false. Enabled if only keep-history is true.
+* **committer-from-gh**: Optional, defaults to false. Allows to use token's owner name and email for commit. Overrides `email` and `name` options.
 * **verbose**: Optional, be verbose about internal steps, defaults to false.
 * **local-dir**: Directory to push to GitHub Pages, defaults to current.
 * **fqdn**: Optional, no default, sets a main domain for your website.
 * **project-name**: Defaults to fqdn or repo slug, used for metadata.
-* **email**: Optional, comitter info, defaults to deploy@travis-ci.org.
-* **name**: Optional, comitter, defaults to Deployment Bot.
+* **email**: Optional, committer info, defaults to deploy@travis-ci.org.
+* **name**: Optional, committer, defaults to Deployment Bot.
 
 #### Examples:
 
--- a/README.md
+++ b/README.md
@@ -520,6 +520,7 @@ You first need to create an [Atlas account](https://atlas.hashicorp.com/account/
 * **repo**: Repo slug, defaults to current one.
 * **target-branch**: Branch to push force to, defaults to gh-pages.
 * **keep-history**: Optional, create incremental commit instead of doing push force, defaults to false.
+* **allow-empty-commit**: Optional, defaults to false. Enabled if only keep-history is true.
 * **verbose**: Optional, be verbose about internal steps, defaults to false.
 * **local-dir**: Directory to push to GitHub Pages, defaults to current.
 * **fqdn**: Optional, no default, sets a main domain for your website.
--- a/lib/dpl/provider/pages.rb
+++ b/lib/dpl/provider/pages.rb
@@ -20,6 +20,12 @@ module DPL
 
       require 'tmpdir'
 
+      if RUBY_VERSION >= "2.0.0"
+        requires 'octokit', version: '~> 4.6.2'
+      else
+        requires 'octokit', version: '~> 4.3.0'
+      end
+
       experimental 'GitHub Pages'
 
       def initialize(context, options)
@@ -66,7 +72,30 @@ module DPL
         options.fetch(:verbose, false)
       end
 
+      def api  # Borrowed from Releases provider
+        error 'gh-token must be provided for Pages provider to work.' unless @gh_token
+
+        @api ||= Octokit::Client.new(:access_token => @gh_token)
+      end
+
+      def user
+        @user ||= api.user
+      end
+
+      def setup_auth
+        user.login
+      end
+
       def check_auth
+        setup_auth
+
+        unless api.scopes.include? 'public_repo' or api.scopes.include? 'repo'
+          error "Dpl does not have permission to access #{@gh_url} using it. Make sure your token contains the repo or public_repo scope."
+        end
+
+        log "Logged in as @#{user.login} (#{user.name})"
+      rescue Octokit::Unauthorized => exc
+        error "gh-token is invalid. Details: #{exc}"
       end
 
       def needs_key?
--- a/lib/dpl/provider/pages.rb
+++ b/lib/dpl/provider/pages.rb
@@ -25,7 +25,7 @@ module DPL
       def initialize(context, options)
         super
 
-        @build_dir = options[:local_dir] || '.'
+        @build_dir = File.join(Dir.pwd, options[:local_dir] || '.')
         @project_name = options[:project_name] || fqdn || slug
         @target_branch = options[:target_branch] || 'gh-pages'
 
@@ -94,7 +94,7 @@ module DPL
 
       def github_init(target_dir)
         FileUtils.cd(target_dir, :verbose => true) do
-          print_step "Creating a brand new local repo from scratch in dir #{`pwd`}..."
+          print_step "Creating a brand new local repo from scratch in dir #{Dir.pwd}..."
           context.shell "git init" or raise 'Could not create new git repo'
           print_step 'Repo created successfully'
           context.shell "git checkout --orphan '#{@target_branch}'" or raise 'Could not create an orphan git branch'
@@ -103,13 +103,13 @@ module DPL
       end
 
       def github_configure
-        print_step "Configuring git committer to be #{@gh_name} <#{@gh_email}> (workdir: #{`pwd`})"
+        print_step "Configuring git committer to be #{@gh_name} <#{@gh_email}> (workdir: #{Dir.pwd})"
         context.shell "git config user.email '#{@gh_email}'"
         context.shell "git config user.name '#{@gh_name}'"
       end
 
       def github_commit
-        print_step "Preparing to deploy #{@target_branch} branch to gh-pages (workdir: #{`pwd`})"
+        print_step "Preparing to deploy #{@target_branch} branch to gh-pages (workdir: #{Dir.pwd})"
         context.shell "touch \"deployed at `date` by #{@gh_name}\""
         context.shell "echo '#{@gh_fqdn}' > CNAME" if @gh_fqdn
         context.shell 'git add -A .'
@@ -118,7 +118,7 @@ module DPL
       end
 
       def github_deploy
-        print_step "Doing the git push (workdir: #{`pwd`})..."
+        print_step "Doing the git push (workdir: #{Dir.pwd})..."
         unless context.shell "git push#{@git_push_opts} --quiet '#{@gh_remote_url}' '#{@target_branch}':'#{@target_branch}' &>/dev/null"
           error "Couldn't push the build to #{@gh_ref}:#{@target_branch}"
         end
@@ -135,7 +135,7 @@ module DPL
             github_pull_or_init(workdir)
 
             FileUtils.cd(workdir, :verbose => true) do
-              print_step "Copying #{@build_dir} contents to #{workdir} (workdir: #{`pwd`})..."
+              print_step "Copying #{@build_dir} contents to #{workdir} (workdir: #{Dir.pwd})..."
               context.shell "rsync -r --exclude .git --delete '#{@build_dir}/' '#{workdir}'" or error "Could not copy #{@build_dir}."
 
               github_configure
--- a/lib/dpl/provider/pages.rb
+++ b/lib/dpl/provider/pages.rb
@@ -123,7 +123,9 @@ module DPL
 
       def github_deploy
         print_step "Doing the git push..."
-        context.shell "git push#{@git_push_opts} --quiet '#{@gh_remote_url}' '#{@target_branch}':'#{@target_branch}' &>/dev/null"
+        unless context.shell "git push#{@git_push_opts} --quiet '#{@gh_remote_url}' '#{@target_branch}':'#{@target_branch}' &>/dev/null"
+          error "Couldn't push the build to #{@gh_ref}:#{@target_branch}"
+        end
       end
 
       def prepare_dir_tree(dir)
@@ -175,9 +177,7 @@ module DPL
               end
               github_configure
               github_commit
-              unless github_deploy
-                error "Couldn't push the build to #{@gh_ref}:#{@target_branch}"
-              end
+              github_deploy
               context.shell "git status" if @verbose
             end
         end
--- a/lib/dpl/provider/pages.rb
+++ b/lib/dpl/provider/pages.rb
@@ -84,7 +84,7 @@ module DPL
         end
 
         print_step "Trying to clone a single branch #{@target_branch} from existing repo..."
-        unless context.shell "git clone --quiet --branch='#{@target_branch}' --depth=1 '#{@gh_remote_url}' '#{target_dir}' &>/dev/null"
+        unless context.shell "git clone --quiet --branch='#{@target_branch}' --depth=1 '#{@gh_remote_url}' '#{target_dir}' > /dev/null 2>&1"
           # if such branch doesn't exist at remote, init it from scratch
           print_step "Cloning #{@target_branch} branch failed"
           Dir.mkdir(target_dir)  # Restore dir destroyed by failed `git clone`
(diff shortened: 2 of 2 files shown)


lib/dpl/provider/pages.rb
```diff
@@ -44,34 +53,140 @@ def slug
         options.fetch(:repo) { context.env['TRAVIS_REPO_SLUG'] }
       end
 
+      def keep_history
+        options.fetch(:keep_history) { false }
```
jyasskin: https://ruby-doc.org/core-2.2.0/Hash.html#method-i-fetch suggests you can write this as `options.fetch(:keep_history, false)` instead of using a block.

lib/dpl/provider/pages.rb
```diff
@@ -44,34 +53,140 @@ def slug
...
       def needs_key?
         false
       end
 
-      def github_deploy
-        context.shell 'rm -rf .git > /dev/null 2>&1'
-        context.shell "touch \"deployed at `date` by #{@gh_name}\""
-        context.shell 'git init' or raise 'Could not create new git repo'
+      def print_step(msg)
+        $stderr.puts msg if @verbose
```
jyasskin: There's a `log` member in https://github.com/travis-ci/dpl/blob/6c3fddfda1f2a85944c544446b068bac0a77c049/lib/dpl/provider.rb#L248 that you should probably use instead of `$stderr.puts`.

lib/dpl/provider/pages.rb
```diff
@@ -44,34 +53,140 @@ def slug
...
+      end
+
+      def github_pull(target_dir)
+        print_step "Trying to clone a single branch #{@target_branch} from existing repo..."
+        unless context.shell "git clone --quiet --branch='#{@target_branch}' --single-branch '#{@gh_remote_url}' '#{target_dir}' &>/dev/null"
```
jyasskin: Consider [`--depth=1`](https://git-scm.com/docs/git-clone#git-clone---depthltdepthgt) to save time when the gh-pages branch gets long. That'll also imply `--single-branch`.

lib/dpl/provider/pages.rb
```diff
@@ -44,34 +53,140 @@ def slug
...
+          # if such branch doesn't exist at remote, do normal clone and create
+          # a new orphan branch
+          print_step "Cloning #{@target_branch} branch failed"
+          print_step 'Trying to clone the whole repo...'
+          context.shell "git clone --quiet '#{@gh_remote_url}' '#{target_dir}' &>/dev/null" or raise "It looks the repo doesn't exist on remote or is inaccessible"
```
jyasskin: You don't need to clone the whole repository to create a new branch; your `git init` trick from when you weren't saving history works fine.

However, you have to explicitly create `target_dir` in that case, since it was removed by the earlier failed clone.

lib/dpl/provider/pages.rb
```diff
@@ -44,34 +53,140 @@ def slug
...
       end
 
       def push_app
-        Dir.mktmpdir {|tmpdir|
-            FileUtils.cp_r("#{@build_dir}/.", tmpdir)
-            FileUtils.cd(tmpdir, :verbose => true) do
-              unless github_deploy
-                error "Couldn't push the build to #{@gh_ref}:#{@target_branch}"
+        print_step "Starting deployment of #{@target_branch} branch to gh-pages..."
+        print_step "The deployment is configured to preserve the target branch if it exists on remote"
```
jyasskin: This needs to be conditional on `@keep_history` if you keep that parameter.

lib/dpl/provider/pages.rb
```diff
@@ -44,34 +53,140 @@ def slug
...
+            print_step "Assuming #{@target_branch} branch doesn't exist, thus creating orphan one"
+            context.shell "git checkout --orphan '#{@target_branch}'"
+          end
+        end
+      end
+
+      def github_clean
+        context.shell "pwd"
+        print_step 'Purging every existing file from repo...'
+        context.shell "git ls-files -z 2>/dev/null | xargs -0 rm -f 2>/dev/null"  # remove all committed files from the repo
```
jyasskin: It's this complexity that `rsync --exclude .git --delete` avoids. Those arguments tell `rsync` to ignore the `.git` directory and to remove everything else that's not in the source, which should match the behavior you've hand-coded here.

lib/dpl/provider/pages.rb
```diff
@@ -44,34 +53,140 @@ def slug
...
+          context.shell "git clone --quiet '#{@gh_remote_url}' '#{target_dir}' &>/dev/null" or raise "It looks the repo doesn't exist on remote or is inaccessible"
+          FileUtils.cd(target_dir, :verbose => @verbose) do
+            print_step "Assuming #{@target_branch} branch doesn't exist, thus creating orphan one"
+            context.shell "git checkout --orphan '#{@target_branch}'"
+          end
+        end
+      end
+
+      def github_clean
+        context.shell "pwd"
```
jyasskin: Probably hide this behind `@verbose`.

lib/dpl/provider/pages.rb
```diff
@@ -44,34 +53,140 @@ def slug
...
+
+      def github_configure
+        print_step "Configuring git committer to be #{@gh_name} <#{@gh_email}>"
         context.shell "git config user.email '#{@gh_email}'"
         context.shell "git config user.name '#{@gh_name}'"
+      end
+
+      def github_commit
+        print_step "Preparing to deploy #{@target_branch} branch to gh-pages"
+        context.shell "touch \"deployed at `date` by #{@gh_name}\""
```
jyasskin: This line makes `--allow_empty_commit` have no effect, since there will always be a 1-file change. That, the existence of #677, #696, and #697, and the fact that I can't think of a use case for this file that's not satisfied by the commit itself, are why I just removed the line in #715.

lib/dpl/provider/pages.rb
```diff
@@ -44,34 +53,140 @@ def slug
...
+            print_step "Created a temporary working directory #{work}"
+
+            return build, work
+      end
+
+      def prepare_build_dir(build, tmp)
+            FileUtils.cp_r("#{build}/.", tmp)
+            print_step "Copied over build contents to #{tmp}"
+            FileUtils.cd(tmp, :verbose => @verbose) do
+                FileUtils.rm_r '.git', :force => true, :verbose => @verbose # cleanup garbage
```
jyasskin: `rsync` also avoids the need for an extra temporary directory to remove the build .git directory.

lib/dpl/provider/pages.rb
```diff
@@ -44,34 +53,140 @@ def slug
...
+
+            print_step "Entering #{tmp_work_dir}..."
+            FileUtils.cd(tmp_work_dir, :verbose => @verbose) do
+              unless @keep_history
+                github_init
               end
+              github_configure
+              github_commit
+              github_deploy
+              context.shell "git status"
```
jyasskin: Also guard this with `@verbose`.

lib/dpl/provider/pages.rb
webknjaz: Sounds reasonable, I just wanted to stick to stdlib calls assuming that mixing lot's of tech (`shell` + `ruby` + variety of third-party tools) might be dirty and would require additional abstraction layer for the sake of readability.

lib/dpl/provider/pages.rb
webknjaz: oops, I probably lost it during rebase

lib/dpl/provider/pages.rb
webknjaz: I'd say remove: I've added this for debugging of that race condition. It's not really needed.

lib/dpl/provider/pages.rb
webknjaz: Sounds interesting. I was used to consider git porcelain more reliable, but looks you are right.

lib/dpl/provider/pages.rb
webknjaz: @jyasskin I think, I saw this in the one of guides I read while researching prior art. Now that I see lots of people are not satisfied with it I think we can safely remove it. On the other hand, I don't want to mix lots of changes in a single PR.

README.md
```diff
@@ -519,6 +519,9 @@ You first need to create an [Atlas account](https://atlas.hashicorp.com/account/
 * **github-token**: GitHub oauth token with `repo` permission.
 * **repo**: Repo slug, defaults to current one.
 * **target-branch**: Branch to push force to, defaults to gh-pages.
+* **keep-history**: Optional, create incremental commit instead of doing push force, defaults to false.
```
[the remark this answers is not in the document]
webknjaz: It would simplify things, but I prefer features to be switchable

README.md
```diff
@@ -519,6 +519,9 @@ You first need to create an [Atlas account](https://atlas.hashicorp.com/account/
...
+* **allow-empty-commit**: Optional, defaults to false. Enabled if only keep-history is true.
```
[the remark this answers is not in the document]
webknjaz: If there was some change, which won't affect anything in generated web-site it would trigger the build anyway. One may want to track that deployment actually took place and this would also indicate that web-site is up-to-date.
If nothing's changed after running deploy, then the deploy is either broken or changes didn't affect web-site. It might be as well a log of deployments in git and it would explicitly indicate that there were some changes in the main branch, while they didn't affect gh-pages.

lib/dpl/provider/pages.rb
webknjaz: @jyasskin I new about this option and I don't remember why I decided not to use it back then. I'll replace it as suggested.

webknjaz:
@jyasskin plz review again. I've cherry-picked your rsync command and did some refactoring.

lib/dpl/provider/pages.rb
```diff
@@ -44,34 +53,92 @@ def slug
...
+      end
+
+      def github_commit
+        print_step "Preparing to deploy #{@target_branch} branch to gh-pages"
+        context.shell "touch \"deployed at `date` by #{@gh_name}\""
         context.shell "echo '#{@gh_fqdn}' > CNAME" if @gh_fqdn
-        context.shell 'git add .'
-        context.shell "FILES=\"`git commit -m 'Deploy #{@project_name} to #{@gh_ref}:#{@target_branch}' | tail`\"; echo \"$FILES\"; echo \"$FILES\" | [ `wc -l` -lt 10 ] || echo '...'"
-        context.shell "git push --force --quiet 'https://#{@gh_token}@#{@gh_ref}' master:#{@target_branch} > /dev/null 2>&1"
+        context.shell 'git add -A .'
```
jyasskin: I believe you don't need the `-A` anymore in order to add removals to the index.

lib/dpl/provider/pages.rb
```diff
@@ -44,34 +53,92 @@ def slug
...
+        unless @keep_history
+          github_init
+          return
+        end
+
+        print_step "Trying to clone a single branch #{@target_branch} from existing repo..."
+        unless context.shell "git clone --quiet --branch='#{@target_branch}' --depth=1 '#{@gh_remote_url}' '#{target_dir}' &>/dev/null"
+          # if such branch doesn't exist at remote, init it from scratch
+          print_step "Cloning #{@target_branch} branch failed"
+          Dir.mkdir(target_dir)  # Restore dir destroyed by failed `git clone`
```
jyasskin: Because this is the directory created by mktmpdir, it's inside a world-writable directory. A concurrent use of mktmpdir could have taken ownership of the directory between when the clone deleted it and when this line tries to recreate it. That's why my patch makes this a subdirectory of the one created by mktmpdir.

README.md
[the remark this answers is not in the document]
153957: I am in favor if keeping the option to force push (discard history), I do not care which way will be the default, but we have some repositories in which the changes in the gh-pages branch are quite large (binary files) and it is not important to keep the history, so to keep the repo size down discarding history is a great feature.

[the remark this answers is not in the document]
webknjaz: /cc: @BanzaiMan

lib/dpl/provider/pages.rb
webknjaz: done

lib/dpl/provider/pages.rb
webknjaz: @jyasskin the defaults for Git 1.x and 2.x are different.
And taking into account that dpl is also an open-source tool which can be freely run outside of @travis-ci infra (in the env with Git 1.x there), I would insist on keeping `-A`. [Explicit is better than implicit.](https://www.python.org/dev/peps/pep-0020/#the-zen-of-python)

Ref: https://stackoverflow.com/a/26039014/595220

webknjaz:
@jyasskin https://travis-ci.org/webknjaz/dpl-gh-pages/builds/315411455#L743 look at this log. What am I missing here? It also looks like mixed output from three providers, which should've been ran subsequently.

jyasskin:
That TravisCI log exposes your token, which starts with `9a9a57`. You should revoke it. This is another reason to use #695.

I don't know why providers would interleave (I'm not a TravisCI expert and have never used multiple providers.)

The `error: src refspec multi-provider-gh-pages-nonexistent does not match any.` error message happens when you write `git push origin this-branch-doesnt-exist:anything`. I haven't re-read your patch to see exactly why it doesn't exist.

lib/dpl/provider/pages.rb
jyasskin: Ah, `-A` sounds good to me then. Thanks for explaining.

webknjaz:
@jyasskin yeah, I know about the token, thanks :)

Regarding multiple providers, when I started my first PR I didn't know either. But @BanzaiMan pointed it to me and that's why I moved all the work to tmpdir. This also means, that we need to test our implementations against this use case as well.

webknjaz:
@jyasskin turned out that removing `&>/dev/null` and `--quiet` from git commands made it work well.
I don't know why, but `unless context.shell "git clone --branch...` doesn't work with those included.

Any ideas?

webknjaz:
Okay, it seems changing `&>/dev/null` to `> /dev/null 2>&1` fixed everything. I bet the underlying shell (ash?) in @travisci boxes doesn't understand `&>` correctly.

webknjaz:
@jyasskin feel free to test this feature-branch now :)

webknjaz:
@jyasskin @BanzaiMan 

This PR is now production-ready: https://travis-ci.org/webknjaz/dpl-gh-pages/builds/323649670

webknjaz:
@BanzaiMan could you please re-run tests?

BanzaiMan:
Looks good. Thank you!!! <3

Could you update the docs as well? https://github.com/travis-ci/docs-travis-ci-com/blob/master/user/deployment/pages.md

webknjaz:
@BanzaiMan Thanks, travis-ci/docs-travis-ci-com#1635 :)

remyabel:
@webknjaz LGTM

webknjaz:
@BanzaiMan this PR should be ready for merging :)

lib/dpl/provider/pages.rb
```diff
@@ -44,34 +63,134 @@ def slug
...
+
+      def verbose
+        # Achtung! Never verbosify git, since it may expose user's token.
+        options.fetch(:verbose, false)
+      end
+
+      def api  # Borrowed from Releases provider
+        error 'gh-token must be provided for Pages provider to work.' unless @gh_token
+
+        @api ||= Octokit::Client.new(:access_token => @gh_token)
```
webknjaz: (just a side note)
It looks I didn't take into account custom GitHub Entreprise URLs.
Need to check whether `@gh_url` could be bypassed in here.

lib/dpl/provider/pages.rb
```diff
@@ -17,23 +21,38 @@ class Pages < Provider
...
+        @keep_history = !!keep_history
+        @allow_empty_commit = !!allow_empty_commit
+        @committer_from_gh = !!committer_from_gh
+        @verbose = !!verbose
 
         @gh_email = options[:email] || 'deploy@travis-ci.org'
         @gh_name = "#{options[:name] || 'Deployment Bot'} (from Travis CI)"
 
         @gh_ref = "#{@gh_url}/#{slug}.git"
+        @gh_remote_url = "https://#{@gh_token}@#{@gh_ref}"
```
webknjaz: (for future)
We could reuse [`~/.netrc` as described here](https://docs.travis-ci.com/user/private-dependencies/#API-Token) as an additional shield against token exposal to log

lib/dpl/provider/pages.rb
webknjaz: [done via #735]

lib/dpl/provider/pages.rb
webknjaz: Unfortunately, git doesn't support bypassing `--netrc-file` argument to underlying `curl` and I don't want to modify user's `~/.netrc`

PR #720, opened 2017-11-27
[Scalingo] Make the app option mandatory

EtienneM:
For the Scalingo provider, we would like to make the `app` option mandatory to make it less error prone.

Working build:
https://travis-ci.org/EtienneM/sample-ruby-sinatra/builds/307954609

### DIFF
--- a/lib/dpl/provider/scalingo.rb
+++ b/lib/dpl/provider/scalingo.rb
@@ -27,37 +27,7 @@ module DPL
 
       def check_auth
         if @options[:api_key]
-          unless context.shell 'mkdir -p ~/.config/scalingo'
-            error "Couldn't create authentication file."
-          end
-          url = URI.parse('http://api.scalingo.com/v1/users/self')
-          http = Net::HTTP.new(url.host, url.port)
-          request = Net::HTTP::Get.new(url.request_uri)
-          request.basic_auth('', @options[:api_key])
-          request['Accept'] = 'application/json'
-          request['Content-type'] = 'application/json'
-          response = http.request(request)
-          data = {}
-          if File.exist?("#{Dir.home}/.config/scalingo/auth")
-            data = JSON.parse(File.read("#{Dir.home}/.config/scalingo/auth"))
-          end
-          begin
-            user = JSON.parse(response.body)
-          rescue
-            error 'Invalid API token.'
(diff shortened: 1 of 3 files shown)


lib/dpl/provider/scalingo.rb
```diff
@@ -6,57 +6,28 @@
...
       end
 
       def initialize(context, options)
         super
+        raise Error, 'you must specify your app name with `--app`' if options[:app] && options[:app].empty?
+
         @options = options
-        @remote = options[:remote] || "scalingo"
-        @branch = options[:branch] || "master"
+        @remote = "git@scalingo.com:#{options[:app]}.git"
```
BanzaiMan: This makes the `app` option mandatory. And existing deployments without it will fail.

lib/dpl/provider/scalingo.rb
```diff
@@ -6,57 +6,28 @@
...
-          data["auth_config_data"]["api.scalingo.com"]["last_name"] = user["user"]["last_name"]
-          data["auth_config_data"]["api.scalingo.com"]["username"] = user["user"]["username"]
-          data["auth_config_data"]["api.scalingo.com"]["email"] = user["user"]["email"]
-          data["auth_config_data"]["api.scalingo.com"]["first_name"] = user["user"]["first_name"]
-          data["auth_config_data"]["api.scalingo.com"]["auth_token"] = @options[:api_key]
-          data["last_update"] = DateTime.now
-          f = File.open("#{Dir.home}/.config/scalingo/auth", "w+") {
-            |f| f.write(data.to_json)
-          }
+          context.shell "timeout 2 ./scalingo login --api-key #{@options[:api_key]} 2> /dev/null > /dev/null"
```
BanzaiMan: Is this compaction really desirable? It is also hiding STDOUT and STDERR without any chance of troubleshooting. I feel that this is wrong.

svenfuchs:
Since you have reviewed the new implementation in master/v2 I am assuming this PR here is not relevant anymore @EtienneM? 

The `app` option currently defaults to the current repo name, see https://github.com/travis-ci/dpl/blob/master/lib/dpl/providers/scalingo.rb#L12

I am going to close this one here. Feel free to re-open if needed.

PR #725, opened 2017-12-09
Remove forced file creation in GitHub Pages provider

### LINKED ISSUE #696: Remove undesired and forced file creation in GitHub Pages deployment

153957:
Fixes #696

### DIFF
--- a/lib/dpl/provider/pages.rb
+++ b/lib/dpl/provider/pages.rb
@@ -53,7 +53,6 @@ module DPL
 
       def github_deploy
         context.shell 'rm -rf .git > /dev/null 2>&1'
-        context.shell "touch \"deployed at `date` by #{@gh_name}\""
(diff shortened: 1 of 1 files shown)


153957:
Also duplicate of #697

PR #740, opened 2018-01-19
Fix side-effect of broken/changed cwd in GitHub Pages deployment provider

webknjaz:
Make `local-dir` rely on `TRAVIS_BUILD_DIR` env variable if available.
Ref #9047.


webknjaz:
@BanzaiMan this fixes behavior described in https://github.com/travis-ci/travis-ci/issues/9047#issuecomment-359118446

Issue #741, opened 2018-01-24
CodeDeploy provider "Aws::S3::Errors::Forbidden" error with dpl 1.8.46 but not dpl 1.8.45


AdrienAgnel:
Hello, 

I have been experiencing deployment issues since yesterday and it seems to me it is related to the upgrade from dpl 1.8.45 to 1.8.46.

I get the following error in my Travis builds :
![image](https://user-images.githubusercontent.com/21367687/35327022-b1562afe-00f8-11e8-9126-b6be2b13eb38.png)
It occurs during a deployment with CodeDeploy provider.

I managed to reproduce the bug locally, using the command detailed here : https://github.com/travis-ci/dpl#examples-32 (the 'key' option appears to be missing in the doc)

- first step, install dpl 1.8.46 and manually run the deployment command for CodeDeploy provider
  - the command fails with the same error stack as above
- second step, install dpl 1.8.45 and manually run the exact same deployment command for CodeDeploy provider
  - the command is successful and a deployment is correctly created

I have no experience with debugging ruby gems, do you have any idea of what is happening ?

BanzaiMan:
It is complaining about the credentials when we call `#head_object` on S3: https://github.com/travis-ci/dpl/blob/af2c5c6fb330011731937e5c019eee36f7d95f92/lib/dpl/provider/code_deploy.rb#L47. Perhaps your key does not have enough authorization for this? And if so, our code should take this possibility into account.

In the mean time, you can force deployment with version 1.8.45:

```yaml
deploy:
  provider: code_deploy
  edge:
    branch: 1.8.45
  ⋮ # rest
```

effata:
This change broke all our travis deployments, which is very unfortunate. Is there an estimate on when this can be fixed? Since you can't give `HeadObject` without giving `GetObject` on S3, a user must have access to download all objects being deployed to CodeDeploy, which I would argue should not be a necessary dependency.
Thank you for the instruction for a work around, even though I'm not looking forward to manually updating all repositories we're deploying through travis.

BanzaiMan:
@effata Understood. I'll work on a fix.

brdmr:
Just a note, the example needs to be prefixed with a `v` to work. I.e. `branch: v1.8.45`. 
Thanks for the quick workaround and fast response time. 👍

BanzaiMan:
@brdmr Indeed. I edited the comment above. Thank you.

AdrienAgnel:
[...]

PR #742, opened 2018-01-24
Fix CodeDeploy provider when key isn't authorized to run `head_object`

BanzaiMan:
It is possible to deploy CodeDeploy app without authorization to run `head_object`. When we encounter such an error, we should move on and fall back to the default behavior.

This resolves https://github.com/travis-ci/dpl/issues/741.

### DIFF
--- a/lib/dpl/provider/code_deploy.rb
+++ b/lib/dpl/provider/code_deploy.rb
@@ -48,7 +48,7 @@ module DPL
           bucket: option(:bucket),
           key: s3_key
         })
-      rescue Aws::S3::Errors
+      rescue Aws::Errors::ServiceError
         {}
       end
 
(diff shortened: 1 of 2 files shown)


lib/dpl/provider/code_deploy.rb
```diff
@@ -44,23 +44,22 @@ def revision
...
         {
           revision_type: 'S3',
           s3_location: {
             bucket:      option(:bucket),
             bundle_type: bundle_type,
             key:         s3_key,
-            version:     s3info[:version_id],
-            e_tag:       s3info[:etag]
-          }
+          }.merge( revision_version_info )
```
effata: This will not work, and generate the error seen here: https://github.com/travis-ci/dpl/issues/741#issuecomment-360515841
The `version` and `e_tag` parameters have changed name, and there's a lot of other data coming from the s3 head_object call that you don't want.

PR #747, opened 2018-01-26
Use sudo to install setuptools twine wheel

BanzaiMan:


lib/dpl/provider/pypi.rb
```diff
@@ -33,8 +33,8 @@ def skip_upload_docs?
       end
 
       def install_deploy_dependencies
-        unless context.shell "wget -O - https://bootstrap.pypa.io/get-pip.py | python - --no-setuptools --no-wheel && " \
-                             "pip install --upgrade setuptools twine wheel"
+        unless context.shell "wget -O - https://bootstrap.pypa.io/get-pip.py | sudo python - --no-setuptools --no-wheel && " \
```
webknjaz: I can understand why you want to do this, but as a user I would not be comfortable with some script intended to do deploy to mess with my system python (and do _anything under root at all_). IMHO, running it should require user to do sudo themself.

ref. https://github.com/travis-ci/dpl/pull/717#discussion_r164240912

BanzaiMan:
This won't work on EC2.

PR #773, opened 2018-03-09
Access gh_token & gh_remote_url correctly

BanzaiMan:
Resolves https://github.com/travis-ci/travis-ci/issues/9312.


BanzaiMan:
Will be in 1.9.1. I'll keep the `pages-ivar-set` branch for a week or so after the 1.9.1 release, so that the repos with this workaround can be switched off it.

PR #783, opened 2018-03-19
Opsworks: Update application source setting after successful deploy

dazz:
Adding new option to also have the application source updated.

Problem: travis only creates deployment but this won't update the settings where one would see the current deployed version. Setting the flag `update-app-on-success` to `true` updates the settings after the deployment was successful.

```yml
deploy:
  provider: opsworks
    access_key_id: abcdefg
    secret_access_key: ultrA-5ecr3t
    app-id: 11223344-5566-7788-9900-aabbccddeeff
    on:
      branch: master
    wait-until-deployed: true
    update-app-on-success: true
```

An opsworks-user with the permission to deploy and to update-app is required.

See successful run here: https://travis-ci.com/easybib/bib-cd/builds/68564088

### DIFF
--- a/lib/dpl/provider/ops_works.rb
+++ b/lib/dpl/provider/ops_works.rb
@@ -101,9 +101,22 @@ module DPL
         print "\n"
         if deployment[:status] == 'successful'
           log "Deployment successful."
+          return unless options[:update_app_on_success]
+          update_app
         else
           error "Deployment failed."
         end
+      end
+
+      def update_app
+        update_config = {
+          app_id: option(:app_id),
+          app_source: {
+            revision: current_sha,
+          }
+        }
+        opsworks.update_app(update_config)
+        log "Application Source branch/revision setting updated."
       end
 
       def wait_until_deployed(deployment_id)


lib/dpl/provider/ops_works.rb
```diff
@@ -101,11 +101,24 @@ def create_deployment
...
       end
 
+      def update_app
+        update_config = {
+          app_id: option(:app_id),
+          app_source: {
+            revision: current_sha,
+          }
+        }
+        opsworks.update_app(update_config)
```
BanzaiMan: I imagine that this could fail. It would be a good idea to check the result of this call and inform the user accordingly if it should fail.

BanzaiMan:
Also, please update README, and also open an accompanying docs PR for https://travis-ci.org/travis-ci/docs-travis-ci-com.

BanzaiMan:
The test deployment looks good, by the way.

lib/dpl/provider/ops_works.rb
dazz: Yes, this once failed because the user did not have the correct permission:

```
Deploying application
creating deployment {"stack_id":"e4494","app_id":"f9e28","command":{"name":"deploy"},"comment":"Deploy build 806 via Travis CI","custom_json":"{\"deploy\":{\"bibcd\":{\"migrate\":false,\"scm\":{\"revision\":\"6ce403\"}}}}"}
Deployment created: 48b9abce
Deployment successful.
Deploying ........................................................
Already up to date!
HEAD detached at 6ce4031
nothing to commit, working tree clean
Dropped refs/stash@{0} (39a3e5a10165cfe324332610312435856d308b81)
Stopping Deploy, OpsWorks service error: arn:aws:iam::888666777:user/username is not authorized to perform: opsworks:UpdateApp
failed to deploy
```

So from that I assume that if an error happens on the opsworks side the it should be fine as the error message is displayed in the logs.

Do you think different?

dazz:
PR to the README is here: https://github.com/travis-ci/docs-travis-ci-com/pull/1769

lib/dpl/provider/ops_works.rb
BanzaiMan: Ah, this is great.

PR #768, opened 2018-03-09
Multithreaded S3 file upload

konrad-c:
This update changes the push_app functionality to upload files to S3 in a parallel fashion. This code has reduced the deploy time of a project from 35 minutes to 6 minutes when using 10 threads and even when using a conservative 5 threads, a build time of 10 minutes.
There are issues associated with the unit tests which expect file yields which are not yielded with an asynchronous, multi-threaded function.
Additionally other tests which expect uploaded file list to match an array. Since order cannot be guaranteed in multi-threaded functions this may require changes. 
The build does not pass due to the tests incorrectly handling multi-threading. Some collaboration on getting the tests working is needed and welcomed.
Travis CI build link: https://travis-ci.org/konrad-c/dpl/jobs/351088693

PR #794, opened 2018-03-27
Fix incorrect S3 upload directory

konrad-c:
The multithreaded file upload to S3 that @mboudreau and myself worked on (https://github.com/travis-ci/dpl/pull/768) changed the use of chdir to a simple glob in order to stop warnings of using a chdir within a chdir. However this caused the upload path of every file to contain the :local_dir option if had been specified. Comparatively, before our changes this was omitted due to the Dir.chdir.
@BanzaiMan 
Do NOT deploy the previous Multithreaded S3 Upload (https://github.com/travis-ci/dpl/pull/768). This PR fixes this issue and should be used in place.


BanzaiMan:
Hello, there! I believe I fixed this issue with #787. Thanks for following up.

PR #802, opened 2018-04-11
Add the option to pin versions for wheel, twine & setuptools

fgogolli:


BanzaiMan:
Thanks for taking the time to open the PR. What is the motivation for this feature?

fgogolli:
The reason for this PR is that, since the latest version of `wheel` (0.31.0), the METADATA version for has been upgraded to 2.1. This new metadata is incompatible with (our) on-premise PyPi repositories, including Artifactory, as can be seen below:

https://www.jfrog.com/jira/browse/RTFACT-16189

This is posing an issue for us, since `dpl` automatically always uses the latest version of `wheel`, `setuptools` and `twine`, and this makes our package uploads fail when using the PyPi deploy provider in Travis.

Throughout this PR, we are able to pin to specific versions for `wheel`, `setuptools` and `twine` and this allows us to overcome this issue and potential future issues which may arise due to such version incompatibilities.

NOTE: We have also logged this issue with our Travis support contacts and they have requested us to proceed this way.

webknjaz:
@fgogolli LGTM, but would be even better if this PR included test specs

lib/dpl/provider/pypi.rb
```diff
@@ -38,9 +38,30 @@ def pypi_skip_existing_option
...
+        wheel_version = options[:wheel_version] || context.env['WHEEL_VERSION'] || ''
+        if !wheel_version.empty?
+          '=' + wheel_version
+        end
+      end
+
       def install_deploy_dependencies
         unless context.shell "wget -O - https://bootstrap.pypa.io/get-pip.py | python - --no-setuptools --no-wheel && " \
-                             "pip install --upgrade setuptools twine wheel"
+                             "pip install --upgrade setuptools#{pypi_pin_setuptools_version} twine#{pypi_pin_twine_version} wheel#{pypi_pin_wheel_version}"
```
BanzaiMan: Tacking on `*_version` return values to the `pip install` arguments is a bit strange. Instead, name the methods differently; e.g., `pypi_setuptools_arg`, which returns `setuptools` by default, but `setuptools=X.Y.Z` when option is given.

lib/dpl/provider/pypi.rb
```diff
@@ -38,9 +38,30 @@ def pypi_skip_existing_option
           end
       end
 
+      def pypi_pin_setuptools_version
+        setuptools_version = options[:setuptools_version] || context.env['SETUPTOOLS_VERSION'] || ''
```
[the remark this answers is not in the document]
BanzaiMan: Need some processing of the user-supplied option is necessary. Imagine, for example, a string including `;` is given.

And appropriate specs are required.

lib/dpl/version.rb
```diff
@@ -1,3 +1,3 @@
 module DPL
-  VERSION = '1.9.5'
+  VERSION = '1.9.6'
```
BanzaiMan: Don't do this.

webknjaz:
@fgogolli it's okay for me, but go ahead and fix @BanzaiMan's comments

lib/dpl/provider/pypi.rb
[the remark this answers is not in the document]
fgogolli: I have tried to make it possible for users to pin the version they would like either within the `.travis.yml` file or as environment variables. The second option would work better for our use-case, so I wanted to allow both.

WRT user input, I can do a basic checking, but, if the user inputs a weird string, it would fail when trying to install. Of course, this opens an attack surface here, which could allow users to inject things to be pip installed, though.

lib/dpl/provider/pypi.rb
fgogolli: Good idea - will do that.

lib/dpl/version.rb
fgogolli: Okay.

fgogolli:
@BanzaiMan I made the changes you requested, so PTAL. Thanks.

lib/dpl/provider/pypi.rb
```diff
@@ -38,9 +38,36 @@ def pypi_skip_existing_option
           end
       end
 
+      def pypi_setuptools_arg
+        setuptools_version = options[:setuptools_version] || context.env['SETUPTOOLS_VERSION'] || ''
+        if setuptools_version[/\A\d+(?:\.\d+)*\z/]
+          'setuptools=' + setuptools_version
```
webknjaz: mb it's `==`? have you tried this on some test repo?

fgogolli:
Hey @webknjaz - thanks for that tip. However, that was the starting point for this PR. Thus, although I tried that, and, it builds the gem locally, Travis still seems to pull it from upstream when installing it - therefore, my changes don't affect my local deploy, which is why I resorted to having this merged here and then try to deploy again.

P.S. Rebased on master.

webknjaz:
@fgogolli Travis is pulling it from the repo/branch which you point to in your `edge:` setting in `deploy:` section.

BanzaiMan:
Hello, there. Could you test the deployment as described in https://github.com/travis-ci/dpl/blob/master/.github/CONTRIBUTING.md#testing-dpl-in-the-context-of-travis-ci-builds and give the build log URL for it? Also, there should be a spec with non-conforming version string(s).

Thanks.

iriberri:
Hey there! I see that the `get-pip.py` script we use also allows to pass the target pip version as an argument (with `python get-pip.py 'pip<6'`, for example).

pip 10 was released at the beginning of April, and looks like it might have some breaking points with pip 9 (https://stackoverflow.com/questions/25192794/no-module-named-pip-req). Would it make sense to expand the work done in this PR and allow for setting up a specific pip version too?

webknjaz:
@iriberri it's an interesting idea. I would even say they we should allow bypassing extra args, because this script also allows installing other packages, which might be used as a deployment requirements injection point.
Also take into account that there's separate versions of get-pip available for 2.6, 3.2 and 3.3 envs: https://github.com/pypa/get-pip#usage

PR #817, opened 2018-05-14
Configurable timeouts for packgecloud provider

capotej:
We received reports from some of our users that their uploads would occasionally take too long and time out. 

This patch allows the user to configure the read, connect, and write timeouts used in `packagecloud` provider. 

Also, it now passes the `DPL::VERSION` along in the user agent.

Link to working build here: https://travis-ci.org/computology/node-test-package

Let me know if there is anything else I need to do to get this merged. Thanks in advance!


akuzminsky:
this still affects us. can we help somehow to get this merged?

PR #826, opened 2018-06-19
crates.io provider for publishing Rust packages

Nemo157:
Example job: https://travis-ci.com/Nemo157/nemo157-rs/jobs/130213124

This is a trivial first integration. It would be possible to add more options like which registry to push to, but those can be left to a future PR.

### DIFF
--- a/lib/dpl/provider/cargo.rb
+++ b/lib/dpl/provider/cargo.rb
@@ -1,0 +1,22 @@
+require 'json'
+require 'uri'
+
+module DPL
+  class Provider
+    class Cargo < Provider
+      def needs_key?
+        false
+      end
+
+      def check_auth
+        option(:token)
+      end
+
+      def push_app
+        if ! context.shell "cargo publish --token #{option(:token)}"
+          raise Error, "Publish failed"
+        end
+      end
+    end
+  end
+end
--- a/README.md
+++ b/README.md
@@ -18,6 +18,7 @@ Dpl supports the following providers:
 * [BitBalloon](#bitballoon)
 * [Bluemix Cloud Foundry](#bluemix-cloud-foundry)
 * [Boxfuse](#boxfuse)
+* [cargo](#cargo)
 * [Catalyze](#catalyze)
 * [Chef Supermarket](#chef-supermarket)
 * [Cloud 66](#cloud-66)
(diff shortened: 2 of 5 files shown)


Nemo157:
New example job given I rebased a change into the first commit: https://travis-ci.com/Nemo157/nemo157-rs/jobs/130230928

Nemo157:
I actually wonder if this provider should be called `cargo` since crates.io is only one instance of a Rust package registry, @carols10cents might you have an opinion on this?

carols10cents:
@Nemo157 Sorry, I don't have any context here-- what is dpl? What does "provider" mean in the context of dpl?

Nemo157:
@carols10cents, sorry, I probably should have mentioned that. `dpl` is used for the deploy providers for Travis.CI, adding this provider will allow you to add something like

```yaml
deploy:
  provider: cratesio
  token: <encrypted API token>
```

To your `.travis.yml` and have it publish the crate to crates.io. As an example this is very close to the existing [`npm` provider](https://docs.travis-ci.com/user/deployment/npm/).

What I’m wondering about is the name used for the provider, should this be crates.io specific since it’s the canonical registry, or should it be something more general in case it’s pointed at an alternative registry.

carols10cents:
Hm, so there would be additional config if publishing to an alt registry? In that case, i think the provider should be cargo.

lib/dpl/provider/cargo.rb
```diff
@@ -0,0 +1,22 @@
...
+
+module DPL
+  class Provider
+    class Cargo < Provider
+      def needs_key?
+        false
+      end
+
+      def check_auth
+        option(:token)
```
BanzaiMan: Is there a way to check with `cargo` if the provided token is valid? If so, it is best to do it here, so that the problem is detected early.

BanzaiMan:
Thanks for the PR. Could you test again, with the new name?

lib/dpl/provider/cargo.rb
Nemo157: Not currently, I might open an issue and see about adding a command for it to `cargo`.

Nemo157:
https://travis-ci.com/Nemo157/nemo157-rs/jobs/135448496 used the new name.

BanzaiMan:
That's using 1.9.6 code base (https://travis-ci.com/Nemo157/nemo157-rs/jobs/135448496#L594), which I believe was what the June code base was. I do not anticipate any significant problem, but please update your branch and try again. (It should use 1.10.0, which would yield 1.10.1 as the version number for your deployment.)

Nemo157:
Rebased and new job should be running soon: https://travis-ci.com/Nemo157/nemo157-rs/jobs/145942198

BanzaiMan:
Please add a page in https://github.com/travis-ci/docs-travis-ci-com/tree/master/user/deployment as well. Thank you!!

PR #540, opened 2016-12-07
removed the preview flag from the gae provider

travi:
fixes #539

PR #830, opened 2018-06-25
Fix some links after #540

peternewman:


peternewman:
This is now passing Travis, although it was only failing before due to changes unrelated to my commit.

PR #833, opened 2018-07-03
Add support for aws lambda event source

### LINKED ISSUE #543: Provider Lambda - Create event source mapping in deployment
When deploying a lambda with travis, it seems we have to set the trigger in a script in after_deploy, using the AWS CLI `create-event-source-mapping`.

It should great to create it in the deploy lambda.

bhavikkumar:
Resolves #543 - Added the ability to create and update lambda function event sources.

The following builds show the functionality working:
- `Create Event Source` https://travis-ci.com/bhavikkumar/lambda-travis-ci/builds/77953875
- `Update Event Source` https://travis-ci.com/bhavikkumar/lambda-travis-ci/builds/77954754

### DIFF
--- a/lib/dpl/provider/lambda.rb
+++ b/lib/dpl/provider/lambda.rb
@@ -101,6 +101,9 @@ module DPL
 
           log "Created lambda: #{response.function_name}."
         end
+        if options[:event_source_arn]
+          event_source(function_name)
+        end
       rescue ::Aws::Lambda::Errors::ServiceException => exception
         error(exception.message)
       rescue ::Aws::Lambda::Errors::InvalidParameterValueException => exception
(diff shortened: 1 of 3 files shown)


BanzaiMan:
Hello, there. Sorry for the delay, and thanks for the PR.

Is there a good reference to what sort of value these event source values can be? Also, does it make sense to validate these values?

lib/dpl/provider/lambda.rb
```diff
@@ -197,7 +228,19 @@ def default_kms_key_arn
       def function_tags
         options[:function_tags] ? split_string_array_to_hash(options[:function_tags]) : nil
       end
-
+      
+      def batch_size
+        options[:batch_size] ? options[:batch_size] : 1
```
BanzaiMan: https://docs.aws.amazon.com/cli/latest/reference/lambda/create-event-source-mapping.html lists various values for the default value, depending on the back end. 1 seems a poor choice.

lib/dpl/provider/lambda.rb
```diff
@@ -197,7 +228,19 @@ def default_kms_key_arn
...
+      end
+      
+      def starting_position
+        options[:starting_position] ? options[:starting_position] : nil
```
BanzaiMan: This ternary operator seems superfluous. Also, should we not validate the user input here?

lib/dpl/provider/lambda.rb
```diff
@@ -197,7 +228,19 @@ def default_kms_key_arn
...
+      end
+      
+      def starting_position_timestamp
+        options[:starting_position_timestamp] ? Time.parse(options[:starting_position_timestamp]) : nil
```
BanzaiMan: Are we passing a `Time` object to `#create_event_source_mapping`? How does that work? Also, don't we want to validate that `starting_position` option is set to `AT_TIMESTAMP`?

PR #834, opened 2018-07-09
pypi: install packages using --user

ssbarnea:
Avoids failure to install due to lack of sudo access.


webknjaz:
@ssbarnea you'll need to do an example build with your changes.

webknjaz:
https://github.com/travis-ci/dpl/blob/master/.github/CONTRIBUTING.md#testing-dpl-in-the-context-of-travis-ci-builds

lib/dpl/provider/pypi.rb
```diff
@@ -66,7 +66,9 @@ def pypi_wheel_arg
       end
 
       def install_deploy_dependencies
-        unless context.shell "wget -O - https://bootstrap.pypa.io/get-pip.py | python - --no-setuptools --no-wheel && " \
+        # --user likely fails inside virtualenvs but is needed outside to avoid needing sudo.
+        unless context.shell "if [ -z ${VIRTUAL_ENV+x} ]; then export PIP_USER=yes; fi && " \
```
webknjaz: This creates hard to understand mix of languages. I'd defer env checks to Ruby side.

lib/dpl/provider/pypi.rb
```diff
@@ -66,7 +66,9 @@ def pypi_wheel_arg
...
+                             "wget -nv -O - https://bootstrap.pypa.io/get-pip.py | python - --no-setuptools --no-wheel && " \
```
webknjaz: What does `-nv` do?

ssbarnea: It disables the progressbar on wget download (cutting ~20 lines of useless console log).

lib/dpl/provider/pypi.rb
ssbarnea: @webknjaz I totally agree with you here. The only problem is that I do not master ruby at all, so if you give me an extra hint on how to do it, I would not mind doing it. Mainly I need to define this new envvar if the other one is defined.

webknjaz: you can probably check `context.env['VIRTUAL_ENV']`

webknjaz:
@ssbarnea test specs need to be updated.

lib/dpl/version.rb
```diff
@@ -1,3 +1,3 @@
 module DPL
-  VERSION = '1.9.7'
+  VERSION = '1.9.8'
```
BanzaiMan: This is not essential to the PR, so please drop this.

lib/dpl/provider/pypi.rb
BanzaiMan: What is `PIP_USER=yes` supposed to do?

lib/dpl/provider/pypi.rb
```diff
@@ -1,7 +1,7 @@
 module DPL
   class Provider
     class PyPI < Provider
-      DEFAULT_SERVER = 'https://upload.pypi.org/legacy/'
+      DEFAULT_SERVER = false
```
BanzaiMan: What is the reason for not defaulting to this server?

BanzaiMan:
Has this been tested?

BanzaiMan:
If anyone is interested in testing this, use:

```yaml
deploy:
  provider: pypi
  edge:
    source: ssbarnea/dpl
    branch: master
  ⋮ # rest
```

kmdouglass:
Hi,
I tested [the fix described above](https://github.com/travis-ci/dpl/pull/834#issuecomment-434036325). It made it seems to fix the pypi error described in #899 , but I then received a new error:

```
python: can't open file 'setup.py': [Errno 2] No such file or directory
``` 

https://travis-ci.org/kmdouglass/RPi-DeviceAdapters/builds/450472764#L2564

ssbarnea:
@kmdouglass This is yours to investigate, probably you are missing a setup.py in your repository and building the package fails.

I stopped using Travis module due to lack of support from Travis team which seems to ignore bugs for months, if not years and not even test and merge fixed made by others.

In the end I switched to us of twine directly as seen in https://github.com/pycontribs/jira/blob/master/.travis.yml#L85 

I do not have more time to invest on this, feel free to update, create another one.... I will let Travis team close it as they do have the power to update it if they want.

webknjaz:
@kmdouglass you navigate away from the dir containing `setup.py` here, in the previous provider: https://github.com/kmdouglass/RPi-DeviceAdapters/commit/a6dc48657adc8f41fba8414f779371cd5ed544fd#diff-354f30a63fb0907d4ad57269548329e3R27
This is why it fails (travis creates a bash script from your config and doesn't care to `cd` back after steps)

cyplo:
Heya, any news on this one ?

BanzaiMan:
@cyplo Have you tested it? If so, please report the results.

cyplo:
@BanzaiMan - I have just tested it and it works for me - see [here](https://travis-ci.org/cyplo/netlify_deployer/builds/488112452).

lib/dpl/version.rb
BanzaiMan:
```suggestion
  VERSION = '1.9.7'
```

PR #844, opened 2018-07-27
[WIP] Add auth_method option for NPM deployment

iriberri:
I learnt from a Support request that some private registries need a specific authentication method, and our approach to check npm's version to use `_auth` or `_authToken`  does not apply in these cases.

For example, to authenticate with Nexus, you need to use the __auth_ method in your .npmrc file no matter what npm version you have.

This PR aims to make possible to specify an authentication method (`auth` or `authtoken` - maybe there's a better naming for these?) so that folks that need to deploy to Nexus, or somewhere else that works similarly, can do it.

Tests and docs will be coming soon -

### DIFF
--- a/lib/dpl/provider/npm.rb
+++ b/lib/dpl/provider/npm.rb
@@ -49,7 +49,8 @@ module DPL
 
       def npmrc_file_content
         log "NPM version: #{npm_version}"
-        if npm_version =~ /^1/
+        
+        if npm_version =~ /^1/ || option(:auth_method) == "auth"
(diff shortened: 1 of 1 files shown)


svenfuchs:
Hey @iriberri :wave:

I have ported your work on this branch to a new PR here: https://github.com/travis-ci/dpl/pull/1037. I believe this should be equivalent to your work.

If you would like me to amend the commit in order to add your Git user name and email address please let me know. (My commit was based on your work, so I would happily pass authorship for it, but I also did not want to presume.)

Thank you!

PR #847, opened 2018-08-01
Allow a user to set a EB version label in the travis YAML

LorneCurrie:
Allow a user to set the EB version label using a YAML property vs ENV variable.

Relates to #862


BanzaiMan:
Thanks for the PR. #860 appears more complete, so let's continue with that one.

PR #851, opened 2018-08-10
Fix GitHub deployment failure when overwriting assets

djpohly:
The overwrite logic for GitHub Releases first checks to see if an asset with the given name already exists, and if so deletes it to make way for the new upload.  If, for some reason, a match is overlooked, the asset
is not deleted, and the upload errors with 422 Validation Failed, code "already_exists" and field "name".

The "some reason" in this case was that only the first page of assets are returned in-line with the releases() call.  Using release_assets() directly will trigger the paginate logic in Octokit and fix the problem.

Here is an example of a build [failing to find and delete](https://travis-ci.org/djpohly/PKGBUILD/builds/414689202#L561) an existing asset from a release, and a build using the PR branch which [successfully overwrites assets](https://travis-ci.org/djpohly/PKGBUILD/builds/414690940#L560) beyond the first page.

### DIFF
--- a/lib/dpl/provider/releases.rb
+++ b/lib/dpl/provider/releases.rb
@@ -36,9 +36,9 @@ module DPL
           }
         }
         if options[:user] and options[:password]
-          @api ||= Octokit::Client.new(:login => options[:user], :password => options[:password], :connection_options => connection_options)
+          @api ||= Octokit::Client.new(:login => options[:user], :password => options[:password], :auto_paginate => true, :connection_options => connection_options)
         else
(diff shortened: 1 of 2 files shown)


djpohly:
Retitled pull request to make the purpose more clear.  Is there anything more I need to do to get some traction/review on this so it can be merged?

djpohly:
@joepvd or @BanzaiMan: Apologies for the explicit tag, but you seem like the currently active devs, and this is a small, straightforward bugfix which still has no developer replies.  Do I need to do anything to help get this merged, other than having patience? :)

BanzaiMan:
Thank you! This looks reasonable.

Issue #859, opened 2018-08-29
required support of `label` for `elasticbeanstalk` provider


jjuraszek:
One of my builds output produce 2 separate application version. I have to name them with meaningful label. Right now it's impossible as only supported way to do it is with `ELASTIC_BEANSTALK_LABEL` which is global for all deployments. Desired format would be:

```
deploy:
  - provider: elasticbeanstalk
    label: first-label
     ...

  - provider: elasticbeanstalk
    label: second-label
     ...
```

jjuraszek:
I was first :P. But seriously... how long does it take to review those changes by travis?

BanzaiMan:
Hello. If you are interested in this feature, please consider looking at the source code and opening a PR. Thank you!

LorneCurrie:
@jjuraszek once BanzaiMan has labelled the Issue and PR, it is pretty quick. I just looked at your PR and you are missing a test run.  Look at the [CONTRIBUTING](https://github.com/travis-ci/dpl/blob/master/.github/CONTRIBUTING.md) doc as to how to do it: [testing-dpl-in-the-context-of-travis-ci-builds](https://github.com/travis-ci/dpl/blob/master/.github/CONTRIBUTING.md#testing-dpl-in-the-context-of-travis-ci-builds)  then include the Build link in your PR.

BanzaiMan:
Please test https://github.com/travis-ci/dpl/pull/860.

PR #860, opened 2018-08-29
allow label to be specified on EB provider

jjuraszek:
it's related to #859


svenfuchs:
@jjuraszek thank you for the PR!

`--label` has been added on master/v2 in the meantime https://github.com/travis-ci/dpl/blob/master/lib/dpl/providers/elasticbeanstalk.rb#L26

The current master has the next major version `v2` which we are about to release. You can opt into using this version by adding:

```
deploy:
  provider: firebase
  edge:
    branch: master
```

I am going to close this one here. Please feel free to re-open it.

PR #864, opened 2018-09-06
Deis/Hephy: Simplify CLI install URL logic

jschneiderhan:
v2 is the only version of the install script being used since cutting over to the Hephy fork of Deis. This change simplifies the URL logic to return the single URL available. I left the `cli_version` option in case any future logic wants to make use of it, even though this was the only code in the provider using it.


Cryptophobia:
Thanks for this simplification @jschneiderhan! It will be nice to make it simpler for now since Hephy will not support Deis v1 anymore.

PR #871, opened 2018-09-21
Prepend https to s3-endpoint if https? is missing

joepvd:


lib/dpl/provider/s3.rb
```diff
@@ -44,7 +44,13 @@ def s3_options
           credentials: ::Aws::Credentials.new(access_key_id, secret_access_key)
         }
 
-        defaults[:endpoint] = options[:endpoint] if options[:endpoint]
+        if options[:endpoint]
+          if options[:endpoint].start_with?('https://', 'http://')
```
BanzaiMan: Is it understood that S3-compatible endpoint is always given in HTTP?

Instead of using `String#start_with?`, I prefer parsing the URL with `URI.parse`:

```ruby
uri = URI.parse(options[:endpoint])
unless uri.scheme
  log "S3 endpoint does not specify scheme; defaulting to HTTPS"
  uri = URI("https://#{options[:endpoint]}")
end
defaults[:endpoint] = uri.to_s
```

joepvd:
https://travis-ci.org/joepvd/travis-experiment/builds/431507037

PR #877, opened 2018-09-28
Provide `quiet` option to S3 provider

BanzaiMan:
Users may end up uploading thousands of files, which will eat up log 
capacity.

### DIFF
--- a/lib/dpl/provider/s3.rb
+++ b/lib/dpl/provider/s3.rb
@@ -26,7 +26,7 @@ module DPL
 
       def check_app
       end
-      
+
       def needs_key?
         false
       end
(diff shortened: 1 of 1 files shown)


BanzaiMan:
This could be a useful option for other potentially noisy providers, but we'll start from here.

soulim:
@BanzaiMan thanks a lot for adding this option. Have you considered a common option for all providers and not just S3? Then it would allow consistent behaviour for all providers.

If I'm not mistaken it should be enough to add conditional logging in this line: https://github.com/travis-ci/dpl/blob/2f13fcaff5293d7463660321c0a197fa1e62ef58/lib/dpl/provider.rb#L294

Something like this:

```ruby
def log(message)
  $stderr.puts(message) unless options[:quiet]
end
```

BanzaiMan:
@soulim I have not considered it. It could work, but I am afraid that doing so without examining each deployment provider's behavior could make diagnosis hard in the case of problems.

BanzaiMan:
I didn't mean to do that.

BanzaiMan:
See #882 for further discussion.

PR #882, opened 2018-10-04
Provide `quiet` option to S3 provider

BanzaiMan:
Users may end up uploading thousands of files, which will eat up log 
capacity.

Redoes #877.

### DIFF
--- a/lib/dpl/provider/s3.rb
+++ b/lib/dpl/provider/s3.rb
@@ -26,7 +26,7 @@ module DPL
 
       def check_app
       end
-      
+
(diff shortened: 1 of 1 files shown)


svenfuchs:
@BanzaiMan i have now added `--verbose` option in https://github.com/travis-ci/dpl/pull/1064 which is `false` by default. only if `--verbose` was given we'd print out all the file names. Otherwise we'd `print` a series of dots. 

WDYT?

i am going to close this one here as it does not target `master` any more.

Issue #411, opened 2016-02-27
Use .ebignore file when deploying to Elastic Beanstalk


petridw:
Would it be possible to have the Elastic Beanstalk provider use the .ebignore file if it is present when creating the .zip to deploy to Elastic Beanstalk, rather than always using .gitignore?

BanzaiMan:
http://docs.aws.amazon.com/elasticbeanstalk/latest/dg/eb-cli3-configuration.html#eb-cli3-ebignore

Seems reasonable.

aripalo:
Yeah, using `.gitignore` instead of `.ebignore` (as is the AWS default when `.ebignore` exists) messes things up: Basically now any build time files - which are git ignored and kept out of vcs - won't get deployed. 

In short: the Elastic Beanstalk deployment is broken.

aripalo:
As `.ebignore` uses the same syntax `.gitignore`, I decided to google for gem that would handle ignore rule vs. file path matching, but only one I found was https://github.com/nathankleyn/gitignore.rb. Which I didn't manage to get installed, it is a wrapper to Rust code and Github doesn't seem active / popular. 

So as an exercise I tried making one myself. It seems to work, but I'll have to test it more with various diffent .ebignore files. If anyone is interested, **some help with testing would be much appreciated**.

The code can be found here: https://gist.github.com/aripalo/b544370b124b23de8c07f81b38e6721f 

When I'm "done", I'll wrap it as a gem, publish it, implement it witin the Elastic Beanstalk provider [`files_to_pack`-function](https://github.com/travis-ci/dpl/blob/master/lib/dpl/provider/elastic_beanstalk.rb#L105) and make a pull request.

jdub:
This is a pretty serious security issue for anyone who moves their deployment process to Travis without realising this doesn't work. `.ebignore` is where you put all the stuff that needs to be in your git repo, but should not be deployed.

jdub:
Here's a cheesy workaround. **Please be careful:** It will only work with relatively simple `.ebignore` files.

```yaml
  # - Deal with lack of .ebignore support in Travis CI's dpl
  #   https://github.com/travis-ci/dpl/issues/411
  # - Use `git rm` so the removed files don't appear in `git ls-files`
  #   https://github.com/travis-ci/dpl/blob/master/lib/dpl/provider/elastic_beanstalk.rb#L106  
  - grep ^/ .ebignore | while read IGNORE; do git rm -rf ".${IGNORE}"; done
[...]

PR #885, opened 2018-10-23
Read .ebignore when choosing files to include in archive

BanzaiMan:
Resolves https://github.com/travis-ci/dpl/issues/411.

1. If `.ebignore` exists, read that file, and that file only, to choose files to include.
1. If `.ebignore` does not exist, but `.gitignore` does, then use the standard rules to choose files. (See https://git-scm.com/docs/git-ls-files#_exclude_patterns)

The latter is a departure from the current behavior, and may require some care.

### DIFF
--- a/README.md
+++ b/README.md
@@ -378,6 +378,13 @@ For example: `--endpoint=https://nyc3.digitaloceanspaces.com`
  * **bucket_path**: Location within Bucket to upload app to.
  * **only_create_app_version**: only create the app version, don't actually deploy it.
 
+If `.ebignore` file exists, the files matching the patterns in this file will be
+ignored during the application upload.
+Similarly, if `.ebignore` does not exist but `.gitignore` does, the files matching
+the standard `.gitignore` rules will be excluded from the upload.
+See [AWS CLI documentation](https://docs.aws.amazon.com/elasticbeanstalk/latest/dg/eb-cli3-configuration.html)
+for more information.
+
 #### Environment variables:
 
  * **ELASTIC_BEANSTALK_ENV**: Elastic Beanstalk environment name which will be updated. Is only used if `env` option is omitted.
--- a/lib/dpl/provider/elastic_beanstalk.rb
+++ b/lib/dpl/provider/elastic_beanstalk.rb
@@ -106,7 +106,14 @@ module DPL
       end
 
       def files_to_pack
-        `git ls-files -z`.split("\x0")
+        exclude_flag = ''
+        if File.exist? '.ebignore'
+          exclude_flag = ' --exclude-from=.ebignore'
+        elsif File.exist? '.gitignore'
+          exclude_flag = ' --exclude-standard'
+        end
+
+        `git ls-files #{exclude_flag} -z`.split("\x0")
       end
 
       def create_zip


BanzaiMan:
Please test this with:

```yaml
deploy:
  provider: elasticbeanstalk
  edge:
    branch: eb-updates
  ⋮ # rest
```

alexjurkiewicz:
Any chance of this getting merged?

BanzaiMan:
No one has reported the results from tests, so I honestly don't know how effective it is.

svenfuchs:
i've been looking at porting this to current master/v2. i'm not sure this is the behavior we need, but i might be mistaken.

do users expect to only include cached (tracked/committed) files (minus the ones in `.ebignore`) to be included in the zip? or do they also expect to add uncached/untracked files? i would have thought expected the latter.

with our `(skip_)cleanup` mechanism i think looking at all files in the current working dir, including both cached/tracked and uncached/untracked ones might make most sense?

```
# some files
$ echo 1 > 1.txt
$ echo 2 > 2.txt
$ echo 3 > 3.txt
$ echo 4 > 4.txt

# some ignore files
$ echo 1.txt > .gitignore
$ echo -e "2.txt\n3.txt" > .ebignore

$ cat .gitignore
1.txt
$ cat .ebignore
2.txt
3.txt

# commit them, except for 3.txt and 4.txt
$ git init
Initialized empty Git repository in /private/tmp/foo/.git/
$ git add 1.txt 2.txt .gitignore .ebignore
The following paths are ignored by one of your .gitignore files:
1.txt
Use -f if you really want to add them.
$ git commit -m 'commit'
[master (root-commit) 3b74553] commit
 3 files changed, 4 insertions(+)
 create mode 100644 .ebignore
 create mode 100644 .gitignore
 create mode 100644 2.txt
```

this does not show 1.txt because it is in `.gitignore`, but it seems we would have expected it to show `3.txt` and `4.txt`, too

```
$ git ls-files
.ebignore
.gitignore
2.txt

these two seem to be the same, so adding `--exclude-standard` might not be needed:

```
$ git ls-files --exclude-from .gitignore
.ebignore
.gitignore
2.txt

$ git ls-files --exclude-standard
.ebignore
.gitignore
2.txt
```

again, this does not include `4.txt`, but also `2.txt` is in .ebignore, why does it get listed?

```
$ git ls-files --exclude-from .ebignore
.ebignore
.gitignore
2.txt
```

so, it would seem to me that `git ls-files` manual description is a little misleading:

> This merges the file listing in the directory cache index with the actual
> working directory list, and shows different combinations of the two.

insofar as `git ls-files` does not by default mix in the actual working dir list, but only includes the index. 

we'd need to add `--others` in order to also get uncommitted files:

```
$ git ls-files --exclude-from .ebignore --others
1.txt
4.txt
```

and both `--cached` and `--others` in order to get both:

```
$ git ls-files --exclude-from .ebignore --others --cached
1.txt
4.txt
.ebignore
.gitignore
2.txt
```

but again there's `2.txt` in there, too, which is in `.ebignore`, and should be ignored.

**Comment by svenfuchs** [comment]
is this just broken in my git version?

```
$ git init
Initialized empty Git repository in /private/tmp/foo/.git/

$ echo 1 > 1
$ echo 1 > .ignore

$ git add 1
$ git commit -am '1'
[master (root-commit) eeb4d0d] 1
 1 file changed, 1 insertion(+)
 create mode 100644 1

$ git ls-files
1
$ git ls-files --exclude-from .ignore
1
$ git ls-files --exclude-from=.ignore
1
```

svenfuchs:
looks like they agreed that it's pretty broken back in 2010 and then never went to add the `--exclude-cached-files-too-pretty-please` option that was agreed on?

http://git.661346.n2.nabble.com/ls-files-exclude-broken-td5536384.html

alexjurkiewicz:
We use and need this functionality. There are some files we want in our repo but not in the eb zip. For example we compile source JS files into bundles with webpack. The source files are in git but ebignored, and the bundle files are in git ignore but are in the eb zip

svenfuchs:
@alexjurkiewicz right, i absolutely understand. 

i believe my findings about `git ls-files` above explain why the approach used in this PR did not work as expected (e.g. https://github.com/travis-ci/dpl/issues/411#issuecomment-445048702)

see this commit mentioned in http://git.661346.n2.nabble.com/ls-files-exclude-broken-td5536384.html

> b5227d80aee5173bfda6aa43a890d03110b0df26 is the first bad commit 
> commit b5227d80aee5173bfda6aa43a890d03110b0df26 
> Author: Jeff King <[hidden email]> 
> Date:   Mon Oct 12 01:11:57 2009 -0400 
> 
>     ls-files: excludes should not impact tracked files 
>     
>     In all parts of git, .gitignore and other exclude files 
>     impact only how we treat untracked files; they should have 
>     no effect on files listed in the index. 
>     
>     This behavior was originally implemented very early on in 
>     9ff768e, but only for --exclude-from. Later, commit 63d285c 
>     accidentally caused us to trigger the behavior for 
>     --exclude-per-directory. 
>     
>     This patch totally ignores excludes for files found in the 
>     index. This means we are reversing the original intent of 
>     9ff768e, while at the same time fixing the accidental 
>     behavior of 63d285c. This is a good thing, though, as the 
>     way that 9ff768e behaved does not really make sense with the 
>     way exclusions are used in modern git. 
>     
>     Signed-off-by: Jeff King <[hidden email]> 
>     Signed-off-by: Junio C Hamano <[hidden email]>

svenfuchs:
We have discussed this more on Slack, and have come to the conclusion that we need a different approach.

Here's a new PR superseding this one here https://github.com/travis-ci/dpl/pull/1057

Issue #843, opened 2018-07-25
Provider for Modulus should be removed


erikthered:
Company was acquired back in 2014 and no longer offers these service, their domain (modulus.io) is actually owned by a different company now. See https://github.com/travis-ci/docs-travis-ci-com/pull/1944

erikthered:
I'll actually try to fix this one this week!

PR #887, opened 2018-10-23
Remove Modulus providers and all related documentation

erikthered:
See #843


BanzaiMan:
Thanks for the PR, but I think this is a duplicate of #886.

PR #891, opened 2018-10-28
Avoid setting GIT_HTTP_USER_AGENT in `atlas`, `releases`, and `script`

BanzaiMan:


BanzaiMan:
Resolves https://github.com/travis-ci/travis-ci/issues/9905.

iriberri:
@BanzaiMan I'm currently debugging this issue happening in the GitHub Releases deployment too. Specifically, when we run `git fetch --tags` to try detect tags on a non-tag build. Do you think we can extend the fix to this provider too? Thank you!

iriberri:
Just updated my comment above with more details :)

BanzaiMan:
The strategy here is to let the `Provider.new` provide the option to remove `GIT_HTTP_USER_AGENT` env var which is already defined when a `DPL::Provider` object is first instantiated.

Issue #845, opened 2018-07-29
[provider: snap] "Unable to locate package snapd" error while installing deploy dependencies 


slodki:
Travis config:
```yaml
deploy:
- provider: snap
  snap: mmex
  channel: edge
  skip_cleanup: true
  on:
    tags: true
    conndition: $OS = snap && ${#SNAP_TOKEN} != 0
```
Travis [log](https://travis-ci.org/moneymanagerex/moneymanagerex/jobs/407817705):
```
$ rvm $(travis_internal_ruby) --fuzzy do ruby -S gem install dpl
Successfully installed dpl-1.9.8
1 gem installed

Installing deploy dependencies
Successfully installed dpl-snap-1.9.8
1 gem installed
E: Unable to locate package snapd
sudo: snap: command not found
Attemping to login

Preparing deploy
/home/travis/.rvm/rubies/ruby-2.2.7/lib/ruby/2.2.0/open3.rb:193:in `spawn': No such file or directory - snapcraft (Errno::ENOENT)
	from /home/travis/.rvm/rubies/ruby-2.2.7/lib/ruby/2.2.0/open3.rb:193:in `popen_run'
	from /home/travis/.rvm/rubies/ruby-2.2.7/lib/ruby/2.2.0/open3.rb:93:in `popen3'
	from /home/travis/.rvm/rubies/ruby-2.2.7/lib/ruby/2.2.0/open3.rb:252:in `capture3'
	from /home/travis/.rvm/gems/ruby-2.2.7/gems/dpl-snap-1.9.8/lib/dpl/provider/snap.rb:32:in `check_auth'
	from /home/travis/.rvm/gems/ruby-2.2.7/gems/dpl-1.9.8/lib/dpl/provider.rb:183:in `block in deploy'
	from /home/travis/.rvm/gems/ruby-2.2.7/gems/dpl-1.9.8/lib/dpl/cli.rb:41:in `fold'
	from /home/travis/.rvm/gems/ruby-2.2.7/gems/dpl-1.9.8/lib/dpl/provider.rb:182:in `deploy'
	from /home/travis/.rvm/gems/ruby-2.2.7/gems/dpl-1.9.8/lib/dpl/cli.rb:32:in `run'
	from /home/travis/.rvm/gems/ruby-2.2.7/gems/dpl-1.9.8/lib/dpl/cli.rb:7:in `run'
[...]

PR #892, opened 2018-10-29
Run `apt-get update` if command is not found

BanzaiMan:
Resolves https://github.com/travis-ci/dpl/issues/845


svenfuchs:
It seems we do something equivalent in `master/v2`? https://github.com/travis-ci/dpl/blob/master/lib/dpl/ctx/bash.rb#L148-L150

I'm going to close this. Please do re-open if needed.

Issue #792, opened 2018-03-22
GCS provider gives virtually no feedback (postive or negative)


BanzaiMan:
gcs provider does not log any meaningful activity. What file is being uploaded, what errors, if any, are encountered while uploading.

See, for example, https://travis-ci.org/cotsog/travis_ci_prod_regular/builds/356889170 (which uses bogus credentials to upload to a nonexistent bucket).

Furthermore, it uses a deprecated gem, so it is best to rewrite it with https://github.com/GoogleCloudPlatform/google-cloud-ruby.

BanzaiMan:
Looking at google-cloud-ruby docs, it appears that it employs a completely different mechanism to authenticate, namely, it uses a key file. It is not impossible, but probably hard and unreliable long-term, to reconstruct this file from bits of information that the user provides. So it is probably going to require a credential file (encrypted, for sure).

kedaly:
So I'm the user.. We can provide a key file no problem, we already have one encrypted in the project. So how do I attach said credential file to the deploy in the .travis.yml file?

BanzaiMan:
@kedaly The current code is badly out of date, and it'll have to be rewritten to work with the new authentication scheme.

kedaly:
@BanzaiMan I've worked around it with docker containers, so no rush :)

BanzaiMan:
I've started working on the new `google-cloud-storage`-based DPL code in the `gcs-ng` branch. There are some significant API changes, however, so it will need some additional testing and user input.

Based on https://github.com/GoogleCloudPlatform/google-cloud-ruby#cloud-storage-ga, I believe that the credentials will have to be stored in a JSON file, not as an API key (as was the case before). This most likely will require [encrypting the file](https://docs.travis-ci.com/user/encrypting-files). And, if you are not careful, the decrypted file may end up in the bucket.

JamieSinn:
@BanzaiMan - you mentioned above a new branch/testing, is that ready to be used in a beta env?

BanzaiMan:
@JamieSinn No, it is not. Sorry.

JamieSinn:
Let me know if/when it is. Until then we'll be unfortunately moving our CI/CD off of Travis :(

BanzaiMan:
https://github.com/googleapis/google-cloud-ruby/blob/master/google-cloud-storage/README.md#authentication says:
> the Service Account credentials can be specified by providing the path to the JSON file, or the JSON itself, in environment variables.

[...]

PR #916, opened 2018-11-09
Switch GCS provider to use Google's API client

BanzaiMan:
gstore no longer works, so we switch to Google's API client.

Resolves https://github.com/travis-ci/dpl/issues/792.

A test deployment: https://travis-ci.org/BanzaiMan/travis_production_test/builds/453035418


BanzaiMan:
Needs specs.

BanzaiMan:
Actually, `gstore`-based provider is working right now. So replacing it without is going to be a little trickier.

BanzaiMan:
Closing this in favor of a rework. Perhaps we'll keep the current implementation and introduce a new "strategy", much like we do with Heroku, to use the newer, Google-supported gem.

PR #917, opened 2018-11-12
Stop using heroku build result endpoint

dmathieu:
This endpoint has been deprecated for years, and we are looking into removing it.
As such, it is currently browned out and will be permanently removed in 2 weeks.
See https://devcenter.heroku.com/changelog-items/1486

This PR stops using that endpoint, to use the standard build one.

PR #919, opened 2018-11-12
Fix heroku result

simmerz:
Uses the build status instead of the deprecated result

I wasn't able to bundle, and the travis setup appears to try and install the heroku gem from rubygems. Let me know what I need to do to remedy this and I will.

I've tested manually and this works from a built gem


dmathieu:
Duplicate of #917 :smile:

simmerz:
@dmathieu yep! I got there a few minutes after you!

PR #931, opened 2018-11-28
Set a custom user-agent for all heroku API calls

dmathieu:
When deprecating the build result endpoint, we didn't see dpl was still using that endpoint due to the lack of custom User Agent, which led us to fix it a bit late (https://github.com/travis-ci/dpl/pull/917).

This sets a custom user agent, to prevent this kind of issue from happening again.

### DIFF
--- a/gemspec_helper.rb
+++ b/gemspec_helper.rb
@@ -39,7 +39,7 @@ def gemspec_for(provider_name=nil, runtime_dependencies=[])
 
     # set up files
     if provider_name
-      s.files       = `git ls-files`.split("\n").grep(Regexp.new provider_name.to_s)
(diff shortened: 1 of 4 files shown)


BanzaiMan:
@dmathieu Thanks for the PR. Has this been tested and confirmed on your end?

dmathieu:
I haven't tested it on an actual app, and I don't have any repository using Travis as a deployment method. I can set that up tomorrow.

BanzaiMan:
@dmathieu That'll be great. Thanks again!

dmathieu:
I have successfully tested this branch: https://travis-ci.com/dmathieu/foobar/builds/92969247
And I confirm from Heroku's end that we do receive requests with the proper user agent (`travis/0.1.0 dpl/1.10.6`).

PR #945, opened 2019-01-15
Fixes test for multiple snaps error

nonword:
Fixes test asserting a specific error is thrown when pushing multiple
snaps. Previous version of test expected a specific error message, which
assumed `Dir.glob` returned results in a specific order, which is system
dependent. ~This fix uses `Dir.glob` in the test itself.~ This fix uses regex to match the error message more flexibly.

### DIFF
--- a/spec/provider/snap_spec.rb
+++ b/spec/provider/snap_spec.rb
@@ -80,7 +80,7 @@ describe DPL::Provider::Snap do
       provider.options[:snap] = "*.snap"
       expect{provider.push_app}.to raise_error(
         DPL::Error,
-        /^Multiple snaps found matching '\*.snap': /
+        /^Multiple snaps found matching '\*.snap': /)
     end
 
     example "missing channel should default to edge" do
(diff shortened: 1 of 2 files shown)


BanzaiMan:
We should probably not match the error message verbatim. Instead, match with a regular expression, or drop the message completely from the expectation completely.

spec/provider/snap_spec.rb
```diff
@@ -80,7 +80,7 @@
       provider.options[:snap] = "*.snap"
       expect{provider.push_app}.to raise_error(
         DPL::Error,
-        "Multiple snaps found matching '*.snap': #{Dir.glob('*.snap').join(', ')}")
+        /^Multiple snaps found matching '\*.snap': /
```
BanzaiMan: Needs to close the parenthesis which was opened on line 81.

BanzaiMan:
Please do not combine two separate concerns into a single PR. Thanks.

BanzaiMan:
I understand the failing builds, but that should be dealt with separately.

nonword:
That's fair. I can submit two separate PRs, but neither will have passing builds because one depends on the other.. Will that be a problem for merging?

BanzaiMan:
Nope. I understand the situation and evaluate them together.

Issue #946, opened 2019-01-15
Nokogiri requires 2.3, travis build targets 2.2.9


nonword:
Trying to [fix a small bug in the snaps provider spec](https://github.com/travis-ci/dpl/pull/945) (in an effort to fix a bug blocking [a different PR](https://github.com/travis-ci/dpl/pull/943)), I find Travis fails only one of the four jobs:

https://travis-ci.org/travis-ci/dpl/builds/480087775

It seems to happen because the target ruby version is 2.2.9, but [Nokogiri requires 2.3+](https://nokogiri.org/#requirements):

```
Fetching: nokogiri-1.10.1.gem (100%)
ERROR:  Error installing dpl-cloud_files-1.10.7.travis.3141.1.gem:
	nokogiri requires Ruby version >= 2.3.0.
rake aborted!
```

nonword:
Proposed fix in this PR: https://github.com/travis-ci/dpl/pull/945

PR #947, opened 2019-01-16
Fix cloud_files gemspec: nokogiri <1.10 for ruby <2.3

nonword:
In cloud_files gemspec, fix nokogiri to <1.10 for ruby <2.3 to accommodate dependence of Nokogiri 1.10 on ruby2.3+

Fixes: https://github.com/travis-ci/dpl/issues/946

### DIFF
--- a/dpl-cloud_files.gemspec
+++ b/dpl-cloud_files.gemspec
@@ -1,3 +1,7 @@
 require './gemspec_helper'
 
-gemspec_for 'cloud_files', [['net-ssh'], ['mime-types'], ['nokogiri'], ['fog-rackspace']]
+if Gem::Version.new(RUBY_VERSION) < Gem::Version.new("2.3.0")
+  gemspec_for 'cloud_files', [['net-ssh'], ['mime-types'], ['nokogiri', '< 1.10'], ['fog-rackspace']]
+else
+  gemspec_for 'cloud_files', [['net-ssh'], ['mime-types'], ['nokogiri'], ['fog-rackspace']]
+end


dpl-cloud_files.gemspec
```diff
@@ -1,3 +1,7 @@
 require './gemspec_helper'
 
-gemspec_for 'cloud_files', [['net-ssh'], ['mime-types'], ['nokogiri'], ['fog-rackspace']]
+if Gem::Version.new(RUBY_VERSION) < Gem::Version.new("2.3.0")
```
BanzaiMan: This is not really effective.

`*.gemspec` is read only once when the gem is built. This means that the resultant gem would have different requirements based on the Ruby version that built it.

This can be a source of confusion if different people build and publish them.

So please don't do this.

dpl-cloud_files.gemspec
```diff
@@ -1,3 +1,7 @@
...
+  gemspec_for 'cloud_files', [['net-ssh'], ['mime-types'], ['nokogiri', '< 1.10'], ['fog-rackspace']]
```
BanzaiMan: We can drop `net-ssh` and `mime-types`.

dpl-cloud_files.gemspec
nonword: I see. Was just following what I saw in the `chef_supermarket` spec, but perhaps that has the same problem.

BanzaiMan: Indeed, I was naïve back then. chef_supermarket shouldn't do that, either. 😬

PR #957, opened 2019-02-17
Fixed configurable "max_thread" issue within S3 provider

chriselsen:
Fixed attempt to compare string with integer, which previously failed.


AlphaSquadAlgorithms:
PATCH

BanzaiMan:
Thanks!

PR #962, opened 2019-02-19
Set up CI with Azure Pipelines

AlphaSquadAlgorithms:
-lib/dpl/provider
      -'Rake-file'


BanzaiMan:
Thanks, but we have no plan to do this.

Issue #969, opened 2019-03-13
Chef supermarket deploy fails with uninitialized constant


majormoses:
Here is the error message:
```
/home/travis/.rvm/gems/ruby-2.4.3/gems/dpl-chef_supermarket-1.10.7/lib/dpl/provider/chef_supermarket.rb:36:in `check_app': uninitialized constant Chef::CookbookUploader (NameError)
Did you mean?  Chef::CookbookLoader
	from /home/travis/.rvm/gems/ruby-2.4.3/gems/dpl-1.10.7/lib/dpl/provider.rb:188:in `block in deploy'
	from /home/travis/.rvm/gems/ruby-2.4.3/gems/dpl-1.10.7/lib/dpl/cli.rb:41:in `fold'
	from /home/travis/.rvm/gems/ruby-2.4.3/gems/dpl-1.10.7/lib/dpl/provider.rb:186:in `deploy'
	from /home/travis/.rvm/gems/ruby-2.4.3/gems/dpl-1.10.7/lib/dpl/cli.rb:32:in `run'
	from /home/travis/.rvm/gems/ruby-2.4.3/gems/dpl-1.10.7/lib/dpl/cli.rb:7:in `run'
	from /home/travis/.rvm/gems/ruby-2.4.3/gems/dpl-1.10.7/bin/dpl:5:in `<top (required)>'
	from /home/travis/.rvm/gems/ruby-2.4.3/bin/dpl:23:in `load'
	from /home/travis/.rvm/gems/ruby-2.4.3/bin/dpl:23:in `<main>'
```

Link to error: https://travis-ci.org/Netuitive/chef-netuitive/jobs/506030147#L1200-L1209

We are an EOL version of chef here but I went back and I see the class being defined here: https://github.com/chef/chef/blob/v12.9.42/lib/chef/cookbook_uploader.rb#L15

Taking a quick look I don't see what the issue is but maybe when I have some more time I can sit down and figure this out

BanzaiMan:
This is curious. RubyGems is not installing the runtime dependency `chef` https://travis-ci.org/Netuitive/chef-netuitive/jobs/506030147#L1193

This is clearly indicated in https://rubygems.org/gems/dpl-chef_supermarket, and it happens if I do that. https://travis-ci.org/BanzaiMan/travis_production_test/builds/509441953#L528

I am not sure why your build does not.

iamjohnnym:
[...]

PR #978, opened 2019-03-26
Require Chef's libraries

BanzaiMan:
Resolves https://github.com/travis-ci/dpl/issues/969

The issue described in #969 should have been present ever since version 1.9.0 (March 8, 2018), but was not reported until very recently. This PR fixes it by `require`ing appropriate files.

(Specs did not catch this error, because the spec file explicitly `require`d indirect dependencies.)

### DIFF
--- a/dpl-chef_supermarket.gemspec
+++ b/dpl-chef_supermarket.gemspec
@@ -1,10 +1,9 @@
 require './gemspec_helper'
 
 if Gem::Version.new(RUBY_VERSION) < Gem::Version.new("2.3.0")
-    gemspec_for 'chef_supermarket', [['rack'], ['mime-types'], ['net-telnet', '~> 0.1.0'], ['chef', '~> 12.0']]
+    gemspec_for 'chef_supermarket', [['rack'], ['mime-types'], ['net-telnet', '~> 0.1.0'], ['chef', '~> 12.0'], ['openssl']]
 elsif Gem::Version.new(RUBY_VERSION) < Gem::Version.new("2.4.0")
-  gemspec_for 'chef_supermarket', [['rack'], ['mime-types'], ['chef', '~> 13.0']]
+  gemspec_for 'chef_supermarket', [['rack'], ['mime-types'], ['chef', '~> 13.0'], ['openssl']]
 else
-  gemspec_for 'chef_supermarket', [['rack'], ['mime-types'], ['chef', '>= 14']]
+  gemspec_for 'chef_supermarket', [['rack'], ['mime-types'], ['chef', '>= 14'], ['openssl']]
 end
-
--- a/dpl-chef_supermarket.gemspec
+++ b/dpl-chef_supermarket.gemspec
@@ -3,16 +3,15 @@ require './gemspec_helper'
 deps = [
   ['rack'],
   ['mime-types'],
-  ['openssl']
 ]
 
 if Gem::Version.new(RUBY_VERSION) < Gem::Version.new("2.3.0")
     deps << ['net-telnet', '~> 0.1.0'] << ['chef', '~> 12.0'] << ['public_suffix', '< 3.1.0']
     gemspec_for 'chef_supermarket', deps
 elsif Gem::Version.new(RUBY_VERSION) < Gem::Version.new("2.4.0")
-  gemspec_for 'chef_supermarket', (deps << ['chef', '~> 13.0'])
+  gemspec_for 'chef_supermarket', (deps << ['chef', '~> 13.0'] << ['openssl'])
 elsif Gem::Version.new(RUBY_VERSION) < Gem::Version.new("2.5.0")
-  gemspec_for 'chef_supermarket', (deps << ['chef', '~> 14.0'])
+  gemspec_for 'chef_supermarket', (deps << ['chef', '~> 14.0'] << ['openssl'])
 else
-  gemspec_for 'chef_supermarket', (deps << ['chef', '>= 14'])
+  gemspec_for 'chef_supermarket', (deps << ['chef', '>= 14'] << ['openssl'])
 end
--- a/lib/dpl/provider/chef_supermarket.rb
+++ b/lib/dpl/provider/chef_supermarket.rb
@@ -1,4 +1,8 @@
 require 'chef'
+require 'chef/config'
+require 'chef/cookbook_loader'
+require 'chef/cookbook_uploader'
+require 'chef/cookbook_site_streaming_uploader'
 
 module DPL
   class Provider
--- a/lib/dpl/provider/chef_supermarket.rb
+++ b/lib/dpl/provider/chef_supermarket.rb
@@ -37,7 +37,7 @@ module DPL
         # So we assume cookbook path is '..'
         cl = ::Chef::CookbookLoader.new '..'
         @cookbook = cl[cookbook_name]
-        cl.validate_cookbooks
+        ::Chef::CookbookUploader.new(cookbook).validate_cookbooks
       end
 
       def push_app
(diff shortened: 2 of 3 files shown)


lib/dpl/provider/chef_supermarket.rb
```diff
@@ -33,7 +37,7 @@ def check_app
         # So we assume cookbook path is '..'
         cl = ::Chef::CookbookLoader.new '..'
         @cookbook = cl[cookbook_name]
-        ::Chef::CookbookUploader.new(cookbook, '..').validate_cookbooks
+        cl.validate_cookbooks
```
iamjohnnym: In looking at the `Chef::CookbookLoader` object, I'm not seeing `validate_cookbooks' but I do see something that might be similar to it.

https://github.com/chef/chef/blob/master/lib/chef/cookbook_loader.rb#L113

`cookbook_exists?`

However, digging a bit deeper.  It looks like, `Chef::CookbookUploader` contains the method `validate_cookbooks`.

https://github.com/chef/chef/blob/master/lib/chef/cookbook_uploader.rb#L140

iamjohnnym: Perhaps, this might be alright?

`::Chef::CookbookUploader.new(cookbook).validate_cookbooks`

BanzaiMan: Let's try that.

BanzaiMan:
@iamjohnnym https://github.com/travis-ci/dpl/issues/969#issuecomment-477200585

The current issue is the instantiation of the `::Chef::CookbookUploader` object. The code suggests that the authentication information is picked up from `::Chef::Config[:client_key]`, but this key failed to properly load. Your config says client key is in [`.travis/client.pem`](https://github.com/socrata-cookbooks/bastion/blob/f560a691563ec6f0eb71722d0e3dfc397bbe7e73/.travis.yml#L49). Are you sure that the file exists when it gets to deploy? I don't know the content of `.travis/configs.tar.gz`, but unless the tarball contains the `.travis/client.pem` directory, `tar xzf .travis/configs.tar.gz` will not create this file.

majormoses:
regarding this error
> /home/travis/.rvm/gems/ruby-2.4.1/gems/chef-14.11.21/lib/chef/http/authenticator.rb:95:in `initialize': password must be at least 4 bytes (OpenSSL::OpenSSLError)

I have seen that locally and installing `openssl` gem fixed it for me, I think it was fixed in a newer version of chef but I can't recall which version we are currently installing.

BanzaiMan:
Interesting. In that case, the `chef` gem needs to list `openssl` explicitly (but it doesn’t). https://rubygems.org/gems/chef

majormoses:
found the issue I recall seeing a while ago: https://github.com/chef/chef/issues/7383

iamjohnnym:
Ah, that's neat.  In a bit, I'll add a line to update the `openssl` gem to my Travis file and see what it does.

majormoses:
looks like in order to fulfill the dependencies we need to use ruby `>= 2.3` which I think is OK as `<= 2.3` is EOL: https://www.ruby-lang.org/en/downloads/branches/

> Ruby 2.2
> status: eol
> release date: 2014-12-25
> EOL date: 2018-03-31

BanzaiMan:
> looks like in order to fulfill the dependencies we need to use ruby >= 2.3

Hmm. That is not really feasible, as we must ensure that it works with Ruby 2.2 (which is the version we guarantee on our build images).

majormoses:
Then we can only install chef `< 13`, which has been EOL for some time: https://www.chef.io/eol-chef12-and-chefdk1/

Can we have the deployer run a different ruby version than the build step? Could we say this deploy step only works on 2.3 `>=`? Out of curiosity how long do you plan on supporting ruby 2.2? It's been EOL for over a year at this point.

BanzaiMan:
https://github.com/travis-ci/dpl/pull/978#issuecomment-478620014 might not be entirely true. I think we should be using the the latest version up to 2.4.x should be used. So, if you are using 2.3.x in your builds, it should work (even on unsupported Precise and older Mac images).

Has anyone tried this PR recently?

majormoses:
> #978 (comment) might not be entirely true. I think we should be using the the latest version up to 2.4.x should be used. So, if you are using 2.3.x in your builds, it should work (even on unsupported Precise and older Mac images).

awesome

> Has anyone tried this PR recently?

I can give it a try tonight if no one beats me to it

majormoses:
Sorry I did actually try last night but I am having a bunch of issues now that I don't think are related. I will report back when I have fixed those issues.

iamjohnnym:
Sorry all, I've been out but I'm back.  Give me a moment and I'll attempt to run my build again.

iamjohnnym:
Looks like I'm still getting, `/home/travis/.rvm/gems/ruby-2.4.1/gems/chef-14.11.21/lib/chef/http/authenticator.rb:95:in `initialize': password must be at least 4 bytes (OpenSSL::OpenSSLError)`

BanzaiMan:
@iamjohnnym Can I see the build log, please?

iamjohnnym:
Absolutely.  Just kicked it off again:

https://travis-ci.org/socrata-cookbooks/bastion/jobs/511745053

BanzaiMan:
Does it require a newer version of the `openssl` gem, perhaps?

dpl-chef_supermarket.gemspec
```diff
@@ -9,9 +9,9 @@ if Gem::Version.new(RUBY_VERSION) < Gem::Version.new("2.3.0")
     deps << ['net-telnet', '~> 0.1.0'] << ['chef', '~> 12.0'] << ['public_suffix', '< 3.1.0']
     gemspec_for 'chef_supermarket', deps
 elsif Gem::Version.new(RUBY_VERSION) < Gem::Version.new("2.4.0")
-  gemspec_for 'chef_supermarket', (deps << ['chef', '~> 13.0'])
+  gemspec_for 'chef_supermarket', (deps << ['chef', '~> 13.0'] << ['openssl'])
```
majormoses: wont all of these resolve based on the system that builds the spec rather than runtime? If your talking about conditionally wanting to install a gem based on the version of ruby you probably need something ugly like this: https://github.com/sensu-plugins/sensu-plugins-supervisor/pull/22

BanzaiMan: You are correct to point out that there is a mismatch between the dependencies we define here, and the dependencies defined by the published gem.

kpfleming:
I have a need to automatically deploy some cookbooks to Supermarket via Travis, so I've come across this PR and the linked issues. Is there anything I can do to assist with testing this?

svenfuchs:
@kpfleming I am not sure about the status of this PR, but we have a major release dpl 2.0 in the works, which (amongst addressing other issues) hopefully helps with reducing these dependency conflicts a little. 

Would you be interested in helping us test the chef supermarket provider on that version? https://github.com/travis-ci/dpl/blob/dpl-2/CONTRIBUTING_2.md As far as the unit tests tell us it should work, but running an integration test would be helpful.

kpfleming:
I would be happy to help test it. Are there instructions for how to use it in a job? The page you linked seems to be for contributing to dpl-2 itself.

svenfuchs:
@kpfleming ah, you're right. it's buried in the example config, sorry about that.

this should do the trick:

```
deploy:
  - provider: chef-supermarket
    edge:
      branch: dpl-2
    ...
```

BanzaiMan:
I think this PR can be closed in favor of the new v2 work @svenfuchs is doing.

PR #1000, opened 2019-06-04
Clean up deps before checking gems

BanzaiMan:
The 'vendor' directory could interfere with our check by
providing a version of dependent gem that might not be satisfied
otherwise by the specifications provided in the gemspec at hand.


BanzaiMan:
Hmm. Does this mean we should remove the cache while running specs, too? 🤔

PR #1007, opened 2019-06-25
npm label updates

ahmadnassri:
- "npm" should be lower-case
- npm terminology is "token" _(vs. api-key)_ for user familiarity


BanzaiMan:
Thank you!

PR #1013, opened 2019-07-11
port scalingo regions to dpl v2

svenfuchs:


lib/dpl/providers/scalingo.rb
```diff
@@ -8,30 +8,33 @@ class Scalingo < Provider
...
+      opt '--timeout SEC', 'Timeout for Scalingo CLI commands', default: 60
 
       needs :git, :ssh_key
 
-      cmds login_key:   'timeout 2 ./scalingo login --api-token %{api_key} > /dev/null 2>&1',
-           login_creds: 'echo -e \"%{username}\n%{password}\" | timeout 2 ./scalingo login > /dev/null 2>&1',
-           add_key:     './scalingo keys-add dpl_tmp_key %{key}',
-           remove_key:  './scalingo keys-remove dpl_tmp_key',
-           remote_add:  'git remote add %{remote} git@scalingo.com:%{app}.git > /dev/null 2>&1',
+      cmds login_key:   'timeout %{timeout} ./scalingo login --api-token %{api_key} > /dev/null 2>&1',
```
EtienneM:
```suggestion
      cmds login_key:   'timeout %{timeout} ./scalingo login --api-token %{api_token} > /dev/null',
```

We actually need the error output to assist our customers when an issue arise

lib/dpl/providers/scalingo.rb
```diff
@@ -8,30 +8,33 @@ class Scalingo < Provider
...
+           login_creds: 'echo -e \"%{username}\n%{password}\" | timeout %{timeout} ./scalingo login > /dev/null 2>&1',
```
EtienneM:
```suggestion
           login_creds: 'echo -e \"%{username}\n%{password}\" | timeout %{timeout} ./scalingo login > /dev/null',
```

lib/dpl/providers/scalingo.rb
```diff
@@ -8,30 +8,33 @@ class Scalingo < Provider
...
-           remote_add: 'Failed to add the git remote.',
-           push:       'Failed to push the app.',
-           remove_key: 'Failed to remove the ssh key.'
+           remove_key: 'Failed to remove the ssh key.',
+           git_setup:  'Failed to add the git remote.',
+           push:       'Failed to push the app.'
 
       def install
         script :install, assert: true
+        ENV['SCALINGO_REGION'] = region if region?
```
EtienneM:
```suggestion
        ENV['SCALINGO_REGION'] = region if region.empty?
```

svenfuchs: @EtienneM are you sure about this? We'd want to set the env var if `region` is present, right? The CLI args parser defines the method `region?` (which returns `true` if any region was given)

svenfuchs: I am going to merge this, as CI is green now after all. Please let me know if I misunderstood this here.

PR #593, opened 2017-03-02
Include submodule files in files_to_pack

jdub:
By default, `git ls-files` does not include submodule files. Add `--recurse-submodules` to ensure they're included, which seems like a pretty reasonable default.

PR #1029, opened 2019-08-14
elasticbeanstalk: use git ls --recurse-submodules (ports #593)

svenfuchs:


svenfuchs:
This is now superseded by https://github.com/travis-ci/dpl/issues/411, which does not use `git ls-files` any more.

PR #600, opened 2017-03-14
Remove non-printable unicode chars from version description in ElasticBeanstalk payload

joepvd:
Reference https://www.w3.org/TR/xml/#charsets

PR #1030, opened 2019-08-14
elasticbeanstalk: remove non-printable unicode chars from description (ports #600)

svenfuchs:


BanzaiMan:
This PR is not exactly ready, in that, AWS has not indicated which characters are valid in the description. See https://github.com/aws/aws-sdk-ruby/issues/1502.

BanzaiMan:
As long as we document how incomplete this is (i.e., that some characters still may get passed through and cause problems), we can merge this, and improve further over time.

svenfuchs:
good call 👍 i'll add a comment, and point to the ticket you've opened on their side

PR #1031, opened 2019-08-14
add pages --deploy_key (ports #695)

svenfuchs:


lib/dpl/ctx/bash.rb
```diff
@@ -490,6 +490,7 @@ def mv(src, dest)
       # Writes the given content to the given file path
       def write_file(path, content, chmod = nil)
         path = File.expand_path(path)
+        FileUtils.mkdir_p(File.dirname(path))
```
webknjaz: This probably needs to go to a separate PR. It doesn't look related.

lib/dpl/providers/pages.rb
```diff
@@ -12,7 +12,10 @@ class Pages < Provider
       gem 'octokit', '~> 4.14.0'
       gem 'public_suffix', '~> 3.0.3'
 
-      opt '--github_token TOKEN', 'GitHub oauth token with repo permission', required: true, secret: true
+      required :github_token, :deploy_key
```
webknjaz: one of?

lib/dpl/providers/pages.rb
```diff
@@ -142,8 +153,20 @@ def git_status
         shell 'git status'
       end
 
+      def setup_deploy_key
+        path = "~/.dpl/deploy_key"
```
webknjaz: I'd prefer it to be put into `ssh-agent` directly instead of storing in the file system.
In bash, you can do things like
```bash
ssh-agent '
  echo $ssh_key_var | ssh-add -
  ssh -T git@github.com  # check auth
  git ...
'
```
(IIRC the syntax)

lib/dpl/providers/pages.rb
```diff
@@ -174,10 +197,18 @@ def sufficient_scopes?
...
+      def remote_url
+        github_token? ? https_url_with_token : git_url
+      end
+
+      def https_url_with_token
         "https://#{github_token}@#{url}"
       end
 
+      def git_url
+        "git@#{url}.git"
```
webknjaz: This pattern is incorrect. For SSH, Git remote URL looks like:
```
git@github.com:user/repo.git  # with a colon
# or
ssh://git@github.com/user/repo.git  # with a slash
```

webknjaz:
I believe this belongs to the label: https://github.com/travis-ci/dpl/labels/provider%3A%20pages%20%28GitHub%20Pages%29

lib/dpl/ctx/bash.rb
svenfuchs: right, not sure why this ended up here

lib/dpl/providers/pages.rb
svenfuchs: yep https://github.com/svenfuchs/cl#required-options

lib/dpl/providers/pages.rb
svenfuchs: it's using the same strategy that other providers use and share, when using a per-build generated ssh key https://github.com/travis-ci/dpl/blob/master/lib/dpl/provider.rb#L332. i think it would be good to use the same approach for consistency, unless there's a strong reason?

lib/dpl/providers/pages.rb
svenfuchs: 👍

svenfuchs:
Woha, i didn't even think of our labels at all. Now added to this one here.

lib/dpl/providers/pages.rb
svenfuchs: i agree there is an argument to be made that it would be good to not clutter the file system with these keys, for cases where people want to run this locally. although that is not the primary use case for dpl, it would make sense to optimize in that direction. however, i think if we considered that, we'd want to do it for all providers that use ssh keys, and stay with one approach.

spec/dpl/providers/pages_spec.rb
```diff
@@ -88,4 +88,15 @@
...
+
+  describe 'given --deploy_key a2V5Cg==', record: true do
+    let(:args) { |e| args_from_description(e) }
+    it { should have_run '[info] Setting up deploy key in ~/.dpl/deploy_key' }
+    it { should have_run '[info] Setting up git-ssh' }
+    it { should have_run '[info] $ ssh -i ~/.dpl/deploy_key -T git@github.com:travis-ci/dpl.git' }
+    it { should have_run 'ssh -i ~/.dpl/deploy_key -T git@github.com:travis-ci/dpl.git' }
+    it { should have_run '[info] Pushing to github.com/travis-ci/dpl.git' }
+    it { should have_run 'git push --quiet "git@github.com:travis-ci/dpl.git" "gh-pages":"gh-pages" > /dev/null 2>&1' }
+    it { should have_run_in_order }
```
webknjaz: Could you please also add a test that the SSH-key file is missing after the script has terminated?
Cases to cover:
1) successful completion
2) failure
3) SIGTERM
4) SIGINT

lib/dpl/providers/pages.rb
webknjaz: No specific reason other than security considerations, like a missing signal handler — I think it's safe to assume that in the event of SIGINT (Ctrl+C) or SIGTERM (kill) there's no robust cleanup set up that'd remove the key file.

I agree that it's probably okay'ish to keep it as is for consistency. As a future improvement, though, it'd be great if such files never hit the disk which could be implemented via a shared module that would be used across providers...

A few extra thoughts:
1) The script should set the folder privileges to 700 too.
2) IIUC semantically the best place for such folder would be under `$XDG_CACHE_HOME` (which is usually `~/.cache`), i.e. `~/.cache/dpl`. Or maybe even `$XDG_RUNTIME_DIR`. https://specifications.freedesktop.org/basedir-spec/latest/ar01s03.html

svenfuchs:
@webknjaz thanks for the additional input!

I have opened a new ticket for considering a different strategy of dealing with these keys https://github.com/travis-ci/dpl/issues/1052

dpl is not really optimized or recommended for use outside of ephemeral environments that can safely be discarded after dpl has run. there are tons of things it does that might wreck havoc on your machine.

that said, i'll also change `master` to `rm -rf .dpl` before the process terminates, which should address your concern, too.

lib/dpl/providers/pages.rb
svenfuchs: >The script should set the folder privileges to 700 too.

sounds reasonable. i'll look into that.

> IIUC semantically the best place for such folder would be under $XDG_CACHE_HOME (which is usually ~/.cache), i.e. ~/.cache/dpl. Or maybe even $XDG_RUNTIME_DIR. https://specifications.freedesktop.org/basedir-spec/latest/ar01s03.html

hmmm, maybe. we used to use `./.dpl` for keeping these things (i.e. in the build working directory), which was a rather bad choice. i've just moved it to the home dir for the time being.

svenfuchs:
added https://github.com/travis-ci/dpl/commit/c00a791df261956a04fe2ab7fb017d8096ce61d5 to remove the entire dir `~/.dpl`

webknjaz:
Random thought: if only https://docs.travis-ci.com/user/private-dependencies/#user-key was available for everyone, dpl wouldn't have to do tricks to support SSH auth...

svenfuchs:
@webknjaz do you also think we can merge this PR? 

also, if you have a working setup, would you like to give it a try?

webknjaz:
I don't have a test setup now but looks okay.

svenfuchs:
alright 👍 

thanks a lot for all the input, @webknjaz!

PR #875, opened 2018-09-24
Default to creds in ~/.aws/credentials

BanzaiMan:
When neither command-line options nor env
vars are set to provide AWS credentials, default to those defined in 
`~/.aws/credentials`.

This is how aws-sdk-ruby behaves.

Resolves #909.

PR #1040, opened 2019-08-14
allow using config files on AWS providers (ports #875)

### LINKED ISSUE #909: Deploying to S3 should read from ~/.aws/credentials file if access_key_id and secret_access_key are not provided
Hi when I try to deploy with s3 my team does not have the access and secret key. However it is on instance's ~/.aws/credentials file that travis is hosted on. Can we add the feature to read from that file before reading the access_key_id and secret_access_key in the .yml file.

here is the error message I get:
```
Preparing deploy
missing access_key_id
failed to deploy
```

svenfuchs:
This resolves #909

### DIFF
--- a/spec/dpl/providers/codedeploy_spec.rb
+++ b/spec/dpl/providers/codedeploy_spec.rb
@@ -41,60 +41,91 @@ describe Dpl::Providers::Codedeploy do
     match { |*| requests[:get_deployment].any? }
   end
 
-  before { subject.run }
-
   let(:github_revision) { { revisionType: 'GitHub', gitHubLocation: { repository: 'dpl', commitId: 'sha' } } }
   let(:s3_revision) { { revisionType: 'S3', s3Location: { bucket: 'bucket', bundleType: 'zip', version: 'ObjectVersionId', eTag: 'ETag' } } }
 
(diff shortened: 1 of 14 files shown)


BanzaiMan:
As with #875, this should be tested.

spec/dpl/providers/codedeploy_spec.rb
```diff
@@ -41,60 +41,91 @@
...
   end
 
-  describe 'given --wait_until_deployed' do
-    it { should create_deployment revision: github_revision }
-    it { should have_run '[print] Waiting for the deployment to finish ' }
-    it { should get_deployment }
+  describe 'with ~/.aws/config' do
+    let(:args) { |e| %w(--access_key_id id --secret_access_key secret) }
+
+    file '~/.aws/config', <<~str
```
BanzaiMan: `<<~` is not valid in Ruby 2.2.x, so we need a different here-doc marker.

svenfuchs: true. i keep forgetting that 😞

svenfuchs: done!

PR #950, opened 2019-01-23
Bintray: Automatically add artifacts to downloads list 

loicalbertin:
This PR fixes #944 
The idea is to allow files in the Bintray file descriptor to be tagged as they should appear in the download list. This way we can select some files to appear in this list and some not (disabled by default).

The tricky part here is that a version should be published before selecting artifacts to appear in the download list. So this feature is disabled if the `publish` is not set to `true` in the file descriptor.
Linked to the publication another trick is while the API responds instantaneously OK to a version publication it takes actually some time (up to 15s in my tests) to be effective. So I had to implement a retry mechanism.

Tested in real conditions using this bintray account: https://bintray.com/beta/#/loicalbertin/dpl-test/auto-upload/
And this Github project https://github.com/loicalbertin/travis-dpl-bintray-test
travis: https://travis-ci.org/loicalbertin/travis-dpl-bintray-test

I updated the bintray file descriptor in README to document this feature.

Please feel free to adapt/tweak retry parameters.

PR #1045, opened 2019-08-14
bintray: add files to downloads list (ports #950)

svenfuchs:


lib/dpl/providers/bintray.rb
```diff
@@ -112,19 +115,41 @@ def publish_version
...
+      def update_files
+        files.each do |file|
+          info :upload_file, source: file.source, target: file.target
+          update_file(file)
+        end
+      end
+
+      def update_file(file)
+        retrying(max: 10, pause: 5) do
+          put(path(:file_metadata, target: file.target), list_in_downloads: true)
```
loicalbertin: Why does `list_in_downloads` is always `true`? And I think that the request body should be converted into json.
I was expecting something like this:

```suggestion
          put(path(:file_metadata, target: file.target), {list_in_downloads: file.download}.to_json)
```

loicalbertin:
I found several issues with this PR.
I fixed them into #1074.

svenfuchs:
@loicalbertin thanks a lot for your work on this!

https://travis-ci.org/loicalbertin/travis-dpl-bintray-test/builds/573880501#L285 looks to me as if this could be merged as soon as i have resolved the merge conflicts on this branch?

i am assuming the uploaded files actually end up being listed on the downloads list?

loicalbertin:
> i am assuming the uploaded files actually end up being listed on the downloads list?

Yes you can check it directly on [bintray overview for this release](https://bintray.com/beta/#/loicalbertin/dpl-test/auto-upload/0.6?tab=overview) while other files appear under [bintray files view](https://bintray.com/beta/#/loicalbertin/dpl-test/auto-upload/0.6?tab=files)

svenfuchs:
@loicalbertin nice! 

ok, so i'll assume we can merge this then, once the build is green.

loicalbertin:
Just a question with this new version of dpl where is the documentation for the bintray release descriptor file (I think it was previously in the README)?

svenfuchs:
@loicalbertin good question, i guess i have overlooked that.

is there a page on the bintray documentation that we can link to, maybe? that way we can make sure our content does not get out of date, and cl has support for adding links to options

svenfuchs:
wow, you're right, it's here: https://github.com/travis-ci/dpl/blob/v1/README.md#bintray

that's a lot of stuff to put into the help output of a command. i think it would be good to be able to link to bintray's documentation, and possibly add some of this to the provider's `description` on the class body. or just move all this to the Travis CI end user docs?

loicalbertin:
This descriptor is dpl-specific so Travis CI end user docs is probably the right place to store it. [In fact it is already present](https://docs.travis-ci.com/user/deployment/bintray/#descriptor-file-example) the fact is I don't know which branch to use as this dpl v2 is not live yet :thinking:

svenfuchs:
@loicalbertin we've sort of pushed documentation updates back a little, and only kept track of these on an internal todo list. 

we are considering branching the whole `deployments` section into two, as both versions might need to be available for a while, but we didn't want to get started on that until we know how many changes here actually will be.

if you'd like to make a PR for this, just branch off of `master`, and mention it's not supposed to be merged until v2 is out. we can then take care of any conflicts later.

svenfuchs:
also, thanks for all the help on this @loicalbertin! ❤️

loicalbertin:
Thank you for taking time to work on this.
I created the documentation PR travis-ci/docs-travis-ci-com#2488

svenfuchs:
awesome, i have added this to our todo list 👍 ❤️

PR #936, opened 2018-12-13
Add Gleis provider

brigriffin:
This PR adds a new provider for deploying Ruby on Rails apps to the https://gleis.cloud PaaS. Currently it supports the following four options:

* **username** (required): Gleis username.
* **password** (required): Gleis password.
* **app** (optional): Gleis application name. Defaults to git repo's name.
* **verbose** (optional): Be verbose about push app process. Defaults to false.

The RSpec tests are included and all pass.

PR #1056, opened 2019-08-16
add gleis provider (ports #936)

svenfuchs:


svenfuchs:
This PR currently also includes the branch `sf-ey-core`, which has a PR here: https://github.com/travis-ci/dpl/pull/1060. This was needed in order to get rid of the conflicting dependency on `rest_client`

svenfuchs:
@brigriffin it looks like the `gleis` gem uses the Ruby 2.3.0 lonely operator https://travis-ci.com/travis-ci/dpl/jobs/226294022#L412, which breaks on Ruby 2.2.0, which we still try to support in `dpl`. otherwise the tests are green :) (well, the second stage didn't run, but i'm pretty sure those would then pass, too)

would it be possible to replace this with something like `count_i && count_i.between?(1, 10)`?

brigriffin:
@svenfuchs first thank you for taking care of the conflicting dependency on the `rest_client` gem and for the heads up regarding the safe navigation operator not being available in Ruby <2.3.

There is a new release (0.6.0) of the `gleis` gem CLI which was pushed this evening to RubyGems.org and which addresses this so that Ruby 2.2 is now supported too.

I will take care of the deployment integration tests probably tomorrow and get back to you directly.

svenfuchs:
@brigriffin awesome, thank you for addressing this so quickly!

for some reason bundler insists on picking `0.1.0` on Ruby 2.2 only as far as i can tell https://travis-ci.com/travis-ci/dpl/jobs/226294022#L490 ... i'll have a closer look at that.

svenfuchs:
@brigriffin ah, it looks like `0.6.0` depends on Ruby `>= 2.3`. with the fixes you made, could that be relaxed to `>= 2.2`?

brigriffin:
@svenfuchs exactly, that's it: in version `0.1.0` of the `gleis` gem there is no `required_ruby_version` restriction in its gemspec file. After `> 0.1.0` the `required_ruby_version`has been set to `>= 2.3` as you correctly mention.

Version `0.6.1` of the `gleis` gem is now out which lowers the `required_ruby_version` version to `>= 2.2` as suggested. We can now happily support people with an old (deprecated) Ruby 2.2...

svenfuchs:
@brigriffin wow, that was quick :) it seems there are more issues with the tests, but i'm sure i'll get now be able to get this green.

so the only thing that is missing for now is the integration test. btw i have signed up for the beta today, so we can use credentials for our machine account. if that's not something Gleis can support that it's also totally fine if you'd submit a PR with encrypted credentials for another account.

svenfuchs:
one minor, non-critical thing: i believe i shouldn't have to require `thor` separately: https://github.com/travis-ci/dpl/blob/1262512ab2fb52f6c2a04ebb4eecb6435da73266/lib/dpl/providers/gleis.rb#L10. however if i don't, then requiring `gleis/cli/addon.rb` would raise ` uninitialized constant Gleis::CLI::Thor`.  i'd recommend adding `require 'thor'` at the top of that file.

brigriffin:
@svenfuchs great, glad to hear that!

Sure, we totally support you guys having an account on Gleis in order to run dpl's integration tests. It is actually best if you install the `gleis` gem and run a `gleis auth register <email>` instead of using the contact form like that you will be able to chose your own password. Bear with us as account creations are still a manual process (sadly enough human check is still the best check against robots,  spammers and such...).

Regarding the `thor` gem requirement I will have to look at it tomorrow as I don't see right now why one would need to add it specifically into the `gleis/cli/addon.rb` file.  If that's the case then it should be added to all `gleis/cli/*.rb` files I suppose.

brigriffin:
Good news I could find the exact reason behind the minor issue with the `thor` gem raising an `uninitialized constant`. A new version of the `gleis` gem should be released this week.

On the other hand I did have not have time to finish the integration tests and will most probably have to postpone this task for next week. I will of course keep you posted.

svenfuchs:
@brigriffin that's alright, we're not in a hurry (on our side at least :) thanks for the update!

brigriffin:
Just a short note to let you know that the `gleis` gem v0.6.2 was released now and addresses the `uninitialized constant` issue, thanks for reporting it. I will get back in touch with you soon again regarding the int tests.

svenfuchs:
@brigriffin thanks for letting me know!

i have updated our side, and the tests are all happy 👍

svenfuchs:
🍏 integration test https://travis-ci.com/travis-ci/dpl/builds/125965085 ... merging this one.

amazing work, @brigriffin! 🎉

PR #1057, opened 2019-08-16
honor .ebignore for excluding files from the zip

svenfuchs:
This should address https://github.com/travis-ci/dpl/issues/411, and supersedes https://github.com/travis-ci/dpl/pull/885

It uses a Rubygem that ports Python's `PathSpec` (which is what the EB CLI uses) https://github.com/highb/pathspec-ruby.

FWIW the Python implementation in the EB CLI currently is:

```python
def get_ebignore_list():
    location = get_ebignore_location()

    if not os.path.isfile(location):
        return None

    with codecs.open(location, 'r', encoding='utf-8') as f:
        spec = PathSpec.from_lines('gitwildmatch', f)

    ignore_list = [f for f in spec.match_tree(get_project_root())]
    ignore_list.append('.ebignore')

    return ignore_list
```

### DIFF
--- a/.travis/providers/aws-elasticbeanstalk/prepare
+++ b/.travis/providers/aws-elasticbeanstalk/prepare
@@ -46,7 +46,3 @@ File.write '.ebignore', <<~str
 str
 
 File.write 'secret.txt', 'secret'
-
-# system 'git init'
-# system 'git add .'
-# system 'git commit -am "dpl test"'
--- a/.travis/providers/aws-elasticbeanstalk/prepare
+++ b/.travis/providers/aws-elasticbeanstalk/prepare
@@ -38,8 +38,15 @@ File.write '.ebextensions/gem_install_bundler.config', <<~str
         gem install bundler -v 2.0.1
 str
 
-# system 'bundle install --path vendor/bundle'
+File.write '.ebignore', <<~str
+  *
+  !Gemfile
+  !config.ru
+  !.ebextensions/gem_install_bundler.config
+str
 
-system 'git init'
-system 'git add .'
-system 'git commit -am "dpl test"'
+File.write 'secret.txt', 'secret'
+
+# system 'git init'
+# system 'git add .'
+# system 'git commit -am "dpl test"'
--- a/.travis/providers/aws-elasticbeanstalk/travis.yml
+++ b/.travis/providers/aws-elasticbeanstalk/travis.yml
@@ -20,6 +20,7 @@ deploy:
     app: dpl-test
     env: DplTest-env
     bucket_name: travis-dpl-test-elasticbeanstalk
+    debug: true
 
 after_deploy:
   - cd ../..
--- a/lib/dpl/providers/elasticbeanstalk.rb
+++ b/lib/dpl/providers/elasticbeanstalk.rb
@@ -30,7 +30,8 @@ module Dpl
       opt '--wait_until_deployed', 'Wait until the deployment has finished'
       opt '--debug', internal: true
 
-      msgs login: 'Using Access Key: %{access_key_id}'
+      msgs login:   'Using Access Key: %{access_key_id}',
+           zip_add: 'Adding %s'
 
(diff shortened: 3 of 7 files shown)


svenfuchs:
works as expected, according to this integration test:

`ebignore` file: https://github.com/travis-ci/dpl/pull/1057/commits/c146153155547d69bc2e358a9fedcbdf512179bd

```
*
!Gemfile
!config.ru
!.ebextensions/gem_install_bundler.config
```

files added to the zip archive: https://travis-ci.com/travis-ci/dpl/builds/123523677#L185

```
Adding Gemfile
Adding .ebextensions/gem_install_bundler.config
Adding config.ru
```

i.e. it has added all the right files, and ignored the file `secret.txt` https://github.com/travis-ci/dpl/pull/1057/commits/c146153155547d69bc2e358a9fedcbdf512179bd#diff-37e597af0ba50fce6fb4086439964915R48

.travis/providers/aws-elasticbeanstalk/prepare
```diff
@@ -38,8 +38,15 @@ File.write '.ebextensions/gem_install_bundler.config', <<~str
...
+  !config.ru
+  !.ebextensions/gem_install_bundler.config
+str
+
+File.write 'secret.txt', 'secret'
 
-system 'git init'
-system 'git add .'
-system 'git commit -am "dpl test"'
+# system 'git init'
```
BanzaiMan: Do we want to eradicate these lines altogether?

.travis/providers/aws-elasticbeanstalk/travis.yml
```diff
@@ -20,6 +20,7 @@ deploy:
     app: dpl-test
     env: DplTest-env
     bucket_name: travis-dpl-test-elasticbeanstalk
+    debug: true
```
BanzaiMan: Is this needed now?

spec/support/now.rb
```diff
@@ -0,0 +1,17 @@
...
+      base.before { allow(Time).to receive(:now).and_return(now) }
+    end
+  end
+end
+
+# https://stackoverflow.com/questions/44259104/rubyzip-undefined-method-to-binary-dos-time-for
+require 'zip'
+
+Zip::DOSTime.instance_eval do
+  def now; Zip::DOSTime.new(); end # ugh.
```
BanzaiMan: What is `ugh` for?

.travis/providers/aws-elasticbeanstalk/prepare
svenfuchs: yep, can do

.travis/providers/aws-elasticbeanstalk/travis.yml
svenfuchs: yes, it makes the provider output the file names/paths added to the zip archive. i don't think we want to do this by default, and i've made the option `internal`, so it won't show up in the readme/help output

lib/dpl/providers/elasticbeanstalk.rb
```diff
@@ -155,6 +172,10 @@ def bucket
       def eb
         @eb ||= Aws::ElasticBeanstalk::Client.new(retry_limit: 10)
       end
+
+      def debug(*args)
```
BanzaiMan: Shouldn't this be more general (e.g., use logger with various log levels, and `debug` is just a synonym for `log_level=debug`)?

spec/support/now.rb
svenfuchs: it's a horrible monkey patch that is needed because rubyzip does something wild that breaks creating the zip file when you stub out `Time.now` in tests. there are a couple open issues on the repo about this

lib/dpl/providers/elasticbeanstalk.rb
svenfuchs: that's a cool idea. would require some work going through other providers that use this same technique.

Issue #829, opened 2018-06-25
GAE deploy multiple configs simultaneously


peternewman:
I've got an issue that's relatively similar to https://github.com/travis-ci/dpl/issues/673 but for a different dpl provider.

I'd like to deploy both an app and an index.yaml file (for Datastore index config) via dpl, as mentioned here:
https://cloud.google.com/sdk/gcloud/reference/app/deploy

And in their example
`gcloud app deploy ~/my_app/app.yaml ~/my_app/another_service.yaml`

It looks like it's just these few lines would need changing:
https://github.com/travis-ci/dpl/blob/master/lib/dpl/provider/gae.rb#L64-L66
https://github.com/travis-ci/dpl/blob/master/lib/dpl/provider/gae.rb#L85

But my Ruby isn't that hot.

I'm currently doing the following:
```
      deploy:
        - provider: gae
          keyfile: my-project-id-key.json
          project: my-project-id
          skip_cleanup: true
          on:
            all_branches: true
        - provider: gae
          keyfile: my-project-id-key.json
          project: my-project-id
          config: index.yaml
          skip_cleanup: true
          on:
            all_branches: true
```

Which works, but means a lot of my config is needlessly repeated. I'd ideally like to just do something like this instead:
```
      deploy:
        provider: gae
        keyfile: my-project-id-key.json
        project: my-project-id
        config:
         - app.yaml
         - index.yaml
        skip_cleanup: true
        on:
          all_branches: true
```

ain:
I just stumbled upon the same thing today. Would be nice to follow `gcloud`'s own logic for the provider.

AustinDizzy:
I too stumbled over it after thinking it supported multiple deployables out-of-the-box and began getting build errors. Just fixed this in #849 and am awaiting the merge now.

peternewman:
This hasn't been fixed and is still applicable.

It's blocked on testing/resolving this comment:
https://github.com/travis-ci/dpl/pull/849#issuecomment-421388516

svenfuchs:
I have changed the `gae` provider to accept multiple config paths in this PR https://github.com/travis-ci/dpl/pull/1067.

@peternewman @ain @AustinDizzy It would be great if you could try this out, and let me know on the PR.

I am going to close this issue here. Please feel free to re-open if needed.

peternewman:
> I have changed the `gae` provider to accept multiple config paths in this PR #1067.
> 
[...]

PR #1067, opened 2019-08-19
gae allow multiple configs (see #829)

svenfuchs:
Can be tested using:

```yaml
deploy:
  provider: gae
  edge:
    branch: sf-gae-configs
⋮
```

### DIFF
--- a/lib/dpl/providers/gae.rb
+++ b/lib/dpl/providers/gae.rb
@@ -15,7 +15,7 @@ module Dpl
 
       opt '--project ID', 'Project ID used to identify the project on Google Cloud', required: true
       opt '--keyfile FILE', 'Path to the JSON file containing your Service Account credentials in JSON Web Token format. To be obtained via the Google Developers Console. Should be handled with care as it contains authorization keys.', default: 'service-account.json'
-      opt '--config FILE', 'Path to your service configuration file', default: 'app.yaml'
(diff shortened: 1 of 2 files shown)


BanzaiMan:
Hmmm. Does this work? I don't see anything working differently when `config` is an array. We accept it, but no real code change?

svenfuchs:
@BanzaiMan yes, it does, transparently. see this unit test https://github.com/travis-ci/dpl/pull/1067/files#diff-7bd87b66e832346332b0f5371016f49cR19

if we pass in the option multiple times then Cl turns it into an array. if `interpolate` evaluates a key to be an array it joins it using a space (because that's what "just works" for most of these cases, if we need some other separator the would have to be joined in an attribute reader style method)

peternewman:
> if we pass in the option multiple times then Cl turns it into an array. if `interpolate` evaluates a key to be an array it joins it using a space (because that's what "just works" for most of these cases, if we need some other separator the would have to be joined in an attribute reader style method)

See also my comment here ( https://github.com/travis-ci/dpl/issues/829#issuecomment-524553764 ), are you saying @svenfuchs  given the multiple times bit I should actually be doing:
```
          config: app.yaml
          config: index.yaml
```

That feels rather different to how the rest of Travis works to me.

Issue #155, opened 2014-07-08
support github release :body


mockitoguy:
Hey guys,

Thanks for awesome tooling! Please consider adding support for :body when creating release in github.

http://rdoc.info/gems/octokit/Octokit/Client/Releases#create_release-instance_method
https://github.com/travis-ci/dpl/blob/master/lib/dpl/provider/releases.rb#L74

Cheers :)

terinjokes:
I, likewise, would love to have the ability to set the body.

abitrolly:
Travis allows to pass body to Octokit. The only problem is to figure out how to quote it for command line - see #657.

valeriomazzeo:
@abitrolly I am having exactly the same issue, did you find a solution?

abitrolly:
@valeriomazzeo no solution. But perhaps the problem is in properly escaping newlines in expanded var.

valeriomazzeo:
@abitrolly I made some progress with this and managed to get a successful upload:

```
before_deploy:
  - brew install jshon
  - export BODY=$(jshon -s "$(cat CHANGELOG.md)"))

deploy:
  provider: releases
  skip_cleanup: true
  api_key: ...
  body: ${BODY}
```

`jhson` properly escapes the markdown, I have tried setting that very same string in `Octokit.rb` and it gets uploaded properly and markdown show us correctly.

Although, when running from travis, markdown doesn't show up correctly, I am trying to remove the double quotes now:

`export BODY=$(sed -e 's/^"//' -e 's/"$//' <<<$(jshon -s "$(cat CHANGELOG.md)"))`

but if `travis/dpl/releases` doesn't add them properly when calling the client function, I am afraid it's not going to work.

BanzaiMan:
While @valeriomazzeo's case appears to show that simple spaces may not be much of a problem, as the value specified for `body` is passed through to `dpl` and Octokit, the spaces (including newlines) could be a problem.

Escaping the `body` value may be able to get around this, I am not entirely certain what that would look like (both on the `dpl` command line, as well as on `.travis.yml`).

Perhaps we can have a separate option that reads the body parameter from a different source (e.g., a file).

valeriomazzeo:
Trimming the double quotes works too, but the markdown doesn't show properly.

Using Octokit.rb directly, works as expected:
```
[...]

PR #1069, opened 2019-08-19
rename --body to --release_notes, add --release_notes_file (see #155)

svenfuchs:
Can be tested using:

```yaml
deploy:
  provider: pages
  edge:
    branch: sf-releases-release-notes
⋮
```

### DIFF
--- a/lib/dpl/providers/releases.rb
+++ b/lib/dpl/providers/releases.rb
@@ -24,11 +24,12 @@ module Dpl
       opt '--overwrite', 'Overwrite files with the same name'
       opt '--prerelease', 'Identify the release as a prerelease'
       opt '--release_number NUM', 'Release number (overide automatic release detection)'
+      opt '--release_notes STR', 'Content for the release notes', alias: :body
+      opt '--release_notes_file PATH', 'Path to a file containing the release notes', note: 'will be ignored if --release_notes is given'
       opt '--draft', 'Identify the release as a draft'
       opt '--tag_name TAG', 'Git tag from which to create the release'
       opt '--target_commitish STR', 'Commitish value that determines where the Git tag is created from'
       opt '--name NAME', 'Name for the release'
-      opt '--body BODY', 'Content for the release notes'
       # should this have --github_url, like Pages does?
 
       needs :git
@@ -41,7 +42,7 @@ module Dpl
            skip_existing:        'File %s already exists, skipping.',
            set_tag_name:         'Setting tag_name to %s',
            set_target_commitish: 'Setting target_commitish to %s',
-           missing_file:         '%s does not exist.',
+           missing_file:         'File %s does not exist.',
            not_a_file:           '%s is not a file, skipping.'
 
       cmds git_fetch_tags:       'git fetch --tags'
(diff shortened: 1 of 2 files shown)


BanzaiMan:
I think `body` is an Octokit/API-provided option that gets tied to Releases, so I am not a fan of removing this option.

svenfuchs:
@BanzaiMan it doesn't remove it, but still keeps it as an alias (so people's deploy configs won't break)

but i'm not sure we have to stick with Octokit's nomenclature necessarily? `body` seems like a rather technical detail to me, and "release notes" is what they use in their documentation https://github.com/octokit/octokit.rb/blob/master/lib/octokit/client/releases.rb#L25 (which is where i've grabbed this name from)

i'm not set on `release_notes`, but i think it should be something that works well with a prefix `_file` ... i'd think `body_file` would be just too weird :)

BanzaiMan:
OK. I have no other concerns.

Midnighter:
@svenfuchs I tried to test it with the following demo repo https://github.com/Midnighter/travis-release-notes/ but my config is not recognized and doesn't trigger a job :disappointed: 

Does it only work for `provider: pages`? I thought most people in #155  wanted it for `provider: releases`.

Also doesn't help that there is no clear way to validate a configuration :/ but that's another issue.

svenfuchs:
@Midnighter it seems your config has both a `matrix` and a `jobs` key. these are aliases, and `matrix` will take precedence, so `jobs` is silently ignored, unfortunately. i'd recommend moving `fast_finish: true` into the `jobs` section, and removing `matrix`.

> Does it only work for `provider: pages`?

whoops, no, sorry. bad copy and paste, i've fixed that. these options are available on the `releases` provider https://github.com/travis-ci/dpl/blob/master/lib/dpl/providers/releases.rb#L28

Midnighter:
Thanks for the hint! I got the deploy to run but now the file is not found and I don't get a lot of information from the log to try and fix it.

https://travis-ci.org/Midnighter/travis-release-notes/jobs/579659390

svenfuchs:
@Midnighter that's because the release provider expects you to specify a `file` option in your build config https://github.com/Midnighter/travis-release-notes/blob/c03b6281fc3cbdbc44a32ead08b31a6b27ae4a36/.travis.yml#L35 (see https://docs.travis-ci.com/user/deployment/releases/)

If you meant to upload all files then you can specify:

```
deploy:
  - provider: pages
     file: *
     file_glob: true
```

I have also opened https://github.com/travis-ci/dpl/pull/1093 that would switch this around a little, and make your current config behave the same. I think that would be slightly better UX, but I'm not yet sure if we want to merge that PR.

If you want to give it a try nonetheless, you can use:
```
deploy:
  - provider: pages
     edge:
       branch: sf-releases-file
```

On that branch `file` and `file_glob` would not be required any more, but default to the same behavior.

svenfuchs:
Also curious that you are getting that noisy backtrace on your build log. That actually shouldn't be there.

Midnighter:
Ah, thank you. I tried to create a minimal example but I'm forgetting the important bits :wink: Thanks for your help.

Midnighter:
Works great! Thank you for the PR and your help here :heart: 

See result https://github.com/Midnighter/travis-release-notes/releases

MarkCallow:
Does this work in dpl v1? I can't use dpl v2 dues to #1213.

PR #1073, opened 2019-08-19
Fixes for cloudformation provider

radekl:
Changes:
- Removed requirement for AWS credentials - it will try the default chain if not provided as parameters (allows to use global env variables and saves some repetitions in .travis.yml later on)
- Fixed events printing
- added/fixed template definition (changed method of `template` -> `template_param` because of infinite loop in messages)
- moved error messages from `errs` to `msgs` - `Cl?` doesn't handle `errs` section (yet?)
- Syntax errors fix for EVENT_KEYS

### DIFF
--- a/lib/dpl/providers/cloudformation.rb
+++ b/lib/dpl/providers/cloudformation.rb
@@ -13,8 +13,8 @@ module Dpl
 
       env :aws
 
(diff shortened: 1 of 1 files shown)


svenfuchs:
looks good to me, @radekl. i'll merge this, and we can do any re-working on the dpl repo. i'm going to DM you on Slack

PR #1074, opened 2019-08-19
Fix issues on sf_port_950 branch

loicalbertin:
Fixed several issues with the original branch:

* When updating file metadata the request body should be json-encoded
* When updating file metadata the request should contain a `Content-Type` set to `application/json`
* When retrying there was no delay
* Failing retries was ignored silently
* There was a mistake that lead to upload only artifacts with the `listInDownloads` attribute but try to actually list all artifacts resulting in a `404` error for artifacts without `listInDownloads` 

Tested on https://github.com/loicalbertin/travis-dpl-bintray-test
See https://travis-ci.org/loicalbertin/travis-dpl-bintray-test/builds/573880501

### DIFF
--- a/lib/dpl/providers/bintray.rb
+++ b/lib/dpl/providers/bintray.rb
@@ -99,8 +99,8 @@ module Dpl
       end
 
       def upload_files
-        files.select(&:download).each do |file|
-          info :list_download, path: file.target
+        files.each do |file|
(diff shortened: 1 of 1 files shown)


svenfuchs:
Looks terrific! Thank you, @loicalbertin. I'll merge this.

PR #1075, opened 2019-08-20
Example for file-based cache control in AWS S3

chriselsen:
Example for file-based cache control in AWS S3

### DIFF
--- a/README.md
+++ b/README.md
@@ -477,6 +477,11 @@ Examples:
 
   dpl s3 --access_key_id id --secret_access_key key --bucket bucket
   dpl s3 --access_key_id id --secret_access_key key --bucket bucket --region region --endpoint url
+  
+  File-based cache control settings:
+  
+  dpl s3 --access_key_id id --secret_access_key key --bucket bucket --cache_control "max-age=99999999, no-cache: foo.html, bar.txt, max-age=9999: *.txt"
+  
 ```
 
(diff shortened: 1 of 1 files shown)


svenfuchs:
@chriselsen thanks for the PR!

the README is autogenerated though, and this would get overwritten the next time we run it.

we'd want to add examples to the option definition here: https://github.com/travis-ci/dpl/blob/master/lib/dpl/providers/s3.rb#L33 ... but i'm almost certain adding multiple examples here wouldn't display well.

i still think it would be best to start documenting this in https://github.com/travis-ci/docs-travis-ci-com/blob/master/user/deployment/s3.md

however, your example would not actually work:

```
--cache_control "max-age=99999999, no-cache: foo.html, bar.txt, max-age=9999: *.txt"
```

this would have to be:

```
--cache_control "no-cache: foo.html, bar.txt" 
--cache_control "max-age=9999: *.txt"
--cache_control "max-age=99999999"
```

or in terms of `.travis.yml`

```yaml
deploy:
  provider: s3
  cache_control: 
    - "no-cache: foo.html, bar.txt" 
    - "max-age=9999: *.txt"
    - max-age=99999999
```

only one mapping per option value, specific ones first, less specific/catch-all ones last

svenfuchs:
I am going to close this PR, adding this to the documentation is on the todo list. thanks!

PR #1082, opened 2019-08-22
fix pages committer to use local git info if no github token is present

svenfuchs:


domenic:
Doesn't seem to have quite worked... https://travis-ci.com/w3c/payment-method-manifest/builds/124293344?utm_medium=notification&utm_source=email

I think you want to run the commands before switching branches?

svenfuchs:
@domenic you're right, we need to configure the git user before changing into the working dir.

this should help: https://github.com/travis-ci/dpl/pull/1082/commits/b8848ba2aebefc492caf958df9f41d716f749986. if it doesn't please let me know about your findings, and i can look into it tomorrow.

thanks for all the help!

domenic:
Hmm. Well, this time it said `Configuring git committer to be Domenic Denicola <d@domenic.me>`, but then during the `Run deployment` step, it has `Author: Travis CI User <travis@example.org>`. And, the push failed, although that could easily be a problem with my deploy key setup or similar, so that part I'd want to do more testing and report back on.

domenic:
I guess it is also pretty suspicious that it says `2736 insertions(+)`. That implies the `keep-history` option isn't really working.

svenfuchs:
ok, this took me a little while.

if we want to get them from the local git commit then the committer name and email need to be evaluated before we switch into the working dir, but we do want to run the command `git config` after switching to the working dir, because we also do not want to use the `--global` flag (in case people run this on their local machine). so we need to access and memoize those methods early.

however, the default of `name` and `email` also gets into the way of this, and it's kind of backwards anyway. i have now made it so that:

* if people give `name` and/or `email` then these will be used
* if people give `committer_from_gh` then the name and email associated with the token will be used (i.e. the token needs the scopes `read:user` and `user:email`), this option now requires the option `github_token` to be given
* by default we fall back to the git author name and email from the current git commit

also, if either we get name and email from the GitHub token or from the current git commit then it will also add `(via Travis CI)` to that name (used to be `from Travis CI`, but i think `via` is better, isn't it?

by default https://travis-ci.com/travis-ci/dpl/builds/124359879#L195

```
Author: Sven Fuchs (via Travis CI) <me@svenfuchs.com>
```

with `committer_from_gh` https://travis-ci.com/travis-ci/dpl/builds/124361191#L196

```
Author: Travis CI dpl test (via Travis CI) <me@svenfuchs.com>
```

with `name` and `email` https://travis-ci.com/travis-ci/dpl/builds/124360440#L195

```
Author: Dpl Test <dpl-test@travis-ci.org>
```

svenfuchs:
i think this is a clear improvement over `master`, so i am going to merge this.

@domenic `keep-history` seems to be working though (it now defaults to `true` btw, so you could remove it from your `.travis.yml` file). it would have erased your commits on https://github.com/w3c/payment-method-manifest/commits/gh-pages if it didn't.

Issue #1079, opened 2019-08-22
Deploy to Ceph-based S3 gateway


mfrancis95:
Hi. I'm trying to deploy to a Ceph-based S3 gateway.

Let's say my endpoint is `https://myendpoint.com` and the bucket is `mybucket`. The deployment tries to access `mybucket.myendpoint.com:443` and gets a `Failed to open TCP connection` error because that name doesn't exist. The gateway is not set up to give buckets their own subdomains. Is there any way I can configure the deployment process to upload to just `myendpoint.com:443`?

The actual way in which a bucket is accessed using this gateway would be `https://myendpoint.com/mybucket`, not `https://mybucket.myendpoint.com`.

svenfuchs:
@mfrancis95 thanks for the report!

I am thinking of adding an option `subdomain` that would default to true, but if set to false would make the provider choose `https://[endpoint]/[bucket]` as an upload target.

What do you think?

mfrancis95:
That sounds great. I need the ability to choose just `https://[endpoint]` as the upload target though.

svenfuchs:
>  I need the ability to choose just https://[endpoint] as the upload target though

Oh, ok? I think I'm confused. How does it then know which bucket the upload should go to?

In your original issue description you said:

> The actual way in which a bucket is accessed using this gateway would be https://myendpoint.com/mybucket

... which made me thought you need `https://[endpoint]/[bucket]`, not `https://[endpoint]`.

Could you clarify? Is this just one single large bucket? How does it get to know which bucket you are referring to?

mfrancis95:
The `bucket` variable is still being used. I think what you said is right.

I'm just coming from my limited experience using `boto3` with a Ceph S3 gateway.

svenfuchs:
I think what we are looking for here might be the option `disable_host_prefix_injection`? https://docs.aws.amazon.com/sdkforruby/api/Aws/S3/Client.html

svenfuchs:
@mfrancis95 i have no idea if this is what you are looking for, but i have added https://github.com/travis-ci/dpl/pull/1083 ... would you like to try this out?

svenfuchs:
gonna merge the PR, closing this. feel free to reopen if needed.

PR #1083, opened 2019-08-23
add s3 --host_prefix

svenfuchs:
refers to https://github.com/travis-ci/dpl/issues/1079

can be tested with

```yaml
deploy:
  provider: s3
  force_path_style: true
  edge:
    branch: sf-s3-subdomain
```

### DIFF
--- a/lib/dpl/providers/s3.rb
+++ b/lib/dpl/providers/s3.rb
@@ -25,7 +25,6 @@ module Dpl
       opt '--bucket BUCKET', 'S3 bucket', required: true
       opt '--region REGION', 'S3 region', default: 'us-east-1'
       opt '--endpoint URL', 'S3 endpoint'
-      opt '--host_prefix', 'Whether to use the bucket name as a subdomain', default: true
       opt '--upload_dir DIR', 'S3 directory to upload to'
       opt '--storage_class CLASS', 'S3 storage class to upload as', default: 'STANDARD', enum: %w(STANDARD STANDARD_IA REDUCED_REDUNDANCY)
       opt '--server_side_encryption', 'Use S3 Server Side Encryption (SSE-AES256)'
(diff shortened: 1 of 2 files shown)


mfrancis95:
From `dpl` would it be tested using `--no-host_prefix`?

svenfuchs:
@mfrancis95 yes, but if you put `host_prefix: false` into your `.travis.yml` file then our build script compiler will pass `--no-host_prefix` to `dpl`

mfrancis95:
Just tested it. It looks like it's still trying to deploy to `[bucket].[endpoint]:443`

svenfuchs:
@mfrancis95 i now think that the option we want might be `force_path_style` https://docs.aws.amazon.com/sdkforruby/api/Aws/S3/Client.html

i have tested this with a local ruby script, and it has output the expected endpoint, keeping the bucket name on the path:

```
opening connection to s3.amazonaws.com:443...
opened
starting SSL for s3.amazonaws.com:443...
SSL established, protocol: TLSv1.2, cipher: ECDHE-RSA-AES128-GCM-SHA256
<- "PUT /travis-dpl-test/dpl_test HTTP/1.1\r\nContent-Type:
\r\nAccept-Encoding: ...
...
```
in order to keep this consistent with S3's terminology i have renamed the option on our side, too. so you now want to pass:

```yaml
deploy:
  provider: s3
  force_path_style: true
  edge:
    branch: sf-s3-subdomain
```

wanna give this a try?

mfrancis95:
@svenfuchs that worked! Thank you. This should make the Travis S3 deployment a lot more compatible with S3 gateways out there.

svenfuchs:
@mfrancis95 we'll release the gem as a preview release today or tomorrow, but that will mean you still need to opt in via `edge: true` (no need to specify a branch then)

there's no timeline on releasing a stable version ... we'll first want to see what kind of issues come up.

svenfuchs:
woha. thanks for the report @mfrancis95 

i've just pushed a change that should fix this.

Issue #867, opened 2018-09-18
New Provider: Docker Registry


poulad:
### Feature :whale: 
A new provider to push docker images(build artifact) to a Docker registry such as hub.docker.com.
Multiple images can be published to different registries with Docker Hub being the default registry.

Here is what I have in mind as options:
```yml
deploy:
  docker:
    - source: local-foo
      target: foo/bar:edge
      registry: hub.docker.com # default
      user: foo
      password:
        - secure: ...
    - source: local-foo:latest
      target: bar/web:edge
      registry: registry.heroku.com
      user: foo
      password:
        - secure: ...
```

This should be easy to implement and in fact, I have already done so in JS. I have no experience with Ruby and cannot add it quickly myself.

Here is the script I use:

<https://github.com/Zevere/WebAPI-Borzoo/blob/110ba3d81af3468678dd52abd815d9b79700d32d/ci/deploy/deploy_docker_registry.js#L9-L12>

and the options (JSON) passed to it:
<https://github.com/Zevere/WebAPI-Borzoo/blob/110ba3d81af3468678dd52abd815d9b79700d32d/.travis.yml#L41>

BanzaiMan:
Thanks for opening the issue. Could you review your proposed YAML configuration? It is an array of 10 elements, which I don't think is what you meant.

poulad:
Thanks @BanzaiMan . I edited the proposed YAML configs.

ankon:
Please re-open: We're missing feedback from @BanzaiMan, but the issue isn't "solved already" nor is it "no longer applicable".

chenrui333:
Any update on this, right now I could use something like below:
https://medium.com/vaidikkapoor/managing-open-source-docker-images-on-docker-hub-using-travis-7fd33bc96d65

mfrancis95:
Also would like an update on this.

A provider that at least provides a shorthand for doing something like
`docker login -p $PASSWORD -u $USERNAME && docker push $TAG` would be helpful.

svenfuchs:
Here's a draft for a Docker provider based on the input on this ticket. I'm not sure it's capturing all the major use cases, and would love to get some feedback from you all:

https://github.com/travis-ci/dpl/pull/1091

svenfuchs:
I'm going to close this issue here. Let's continue the conversation on the PR. Feel free to re-open if needed though.

PR #1091, opened 2019-08-30
Add a docker provider (draft)

svenfuchs:


mfrancis95:
Is the image name argument really required?

See https://github.com/travis-ci/dpl/issues/867#issuecomment-524594611.

I've been doing successful deployments just using `docker build -t $TAG` and `docker push $TAG`.

svenfuchs:
@mfrancis95 makes sense. i have removed the requirement.

would you be able to test this out? ideally in a real-world deployment usecase?

```yaml
deploy:
- provider: docker
   edge:
     branch: sf-docker
```

mfrancis95:
I tested it out. When `image` is not included, `docker build . --no-cache -t` gets run without a tag. I think in this case the `target` should be used in its place.

TotoGaz:
Hello,

I’m new to travis-ci and CI/CD, I’ve tried this patch nevertheless.

I guess your `--context` option (defined as the `Path or URL to the Dockerfile context`) merges two docker ideas: the `context` and the `Dockerfile`.
Dockerfile being at the root of the context is only the default behaviour and DPLv2 does not seem to provide any workaround if this is not the case.
Docker provides the possibility to define the Dockerfile we wan to use (through the `--file/-f` option): this option should be available in DPLv2.
Also, since there are so many options in docker build, providing an _“extra”_ way to define options could be considered.

Another thing that bothers me a little is the `image` and `target`.
I am no expert of docker but the way it is used in DPLv2 makes me think `image` like an intermediate temporary argument. Because only `target` will be pushed (and therefore visible).
_This needs to be confirmed_ but only one `dest` info could be used alongside the `docker build -t ${dest} …`, and no need to `docker tag` anymore.
Maybe this is also the reason why @mfrancis95 did try without including `image` (I don’t know :)).

Thanks you!

TotoGaz:
Another point, in my logs I could read
`docker login --username [secure] --password 9Wp*****************`.
Is legit to exhibit the first 3 characters of the password?

mfrancis95:
Might need to use the `--password-stdin` flag.

PR #1097, opened 2019-09-11
do not commit unless the working dir has changes

svenfuchs:


lib/dpl/providers/pages/git.rb
```diff
@@ -96,7 +96,7 @@ def deploy
           git_clone? ? git_clone : git_init
           copy_files
           git_config
-          git_commit
+          git_commit if git_dirty?
           git_push
```
BanzaiMan: Do we need to push if the local repo isn't dirty? In fact, if it's not dirty, do we need to bother setting up git?

svenfuchs: good points ... 🤔

svenfuchs: addressed, i think?

PR #1100, opened 2019-09-13
honor .gitignore if .ebignore is not present

svenfuchs:


lib/dpl/providers/elasticbeanstalk.rb
```diff
@@ -141,10 +141,14 @@ def check_deployment(msgs)
 
       def files
         files = Dir.glob('**/*', File::FNM_DOTMATCH)
-        files = filter(files, '.ebignore') if file?('.ebignore')
+        files = filter(files, ignore_file) if ignore_file
         files
       end
 
+      def ignore_file
+        @ignore_file ||= %w(.ebignore .gitignore).detect { |file| file?(file) }
```
BanzaiMan: Does this need to be memoized? I don't think we should worry too much, but, logically speaking, if the file is initially detected but subsequently deleted somehow, `#filter` would fail.

For this reason, maybe it's better to roll this logic into `#filter` instead.

svenfuchs: ok, moved it to `filter` :)

PR #1114, opened 2019-10-07
NPM: add --dry-run option

saschazar21:
This PR introduces the `--dry-run` CLI parameter as described in the [npm CLI documentation](https://docs.npmjs.com/cli/publish.html#description).

The reason for this addition is, that there is currently no possibility of testing the `npm publish` option without actually uploading the package to the registry.

Due to the fact, that a version number cannot be re-used on npm once a package has been uploaded and released on the registry, it could cause issues in terms of the transparency between tagged versions vs. the actually released packages.


svenfuchs:
@saschazar21 looks good, thank you!

i'll merge this 🎉

PR #1121, opened 2019-10-15
fix: removed separator which caused issues

radekl:
+ Removed irrelevant assume role parameter (not suitable for ci/cd)


radekl:
Need to test if it helps

radekl:
Now it's good. I have tested it with simple template. I guess it's ready to be merged.

Issue #1118, opened 2019-10-11
Allow override of `npm publish`


kripod:
When publishing packages of a monorepo, a custom npm script needs to be used (e.g. `npm run publish`, which would execute `lerna publish from-package --contents pkg` or similar, see [the corresponding issue](https://github.com/pikapkg/pack/issues/1) of pikapkg/pack).

I think that if a `publish` script is present in `package.json`, it should be preferred over the default publishing method.

svenfuchs:
@kripod thanks for opening this!

is our understanding correct that you'd like `dpl npm` use `npm run publish` rather than `npm publish` automatically if the `package.json` contains a `publish` script?

if we did so we are concerned that we'd do so for users who do not want that, so our thinking is that we'd want an additional opt-in option, e.g. `--run_script` (or similar).

how common is it to define a `publish` script in the `package.json` file?

kripod:
> is our understanding correct that you'd like `dpl npm` use `npm run publish` rather than `npm publish` automatically if the `package.json` contains a `publish` script?

Yeah, I proposed it exactly as you've described.

> if we did so we are concerned that we'd do so for users who do not want that, so our thinking is that we'd want an additional opt-in option, e.g. `--run_script` (or similar).

An opt-in feature would be very sensitive, great idea indeed.

> how common is it to define a `publish` script in the `package.json` file?

I'm not sure about it, but it's a viable use-case [for monorepos](https://github.com/lerna/lerna/tree/master/commands/publish#readme).

svenfuchs:
@kripod i have added https://github.com/travis-ci/dpl/pull/1124 which i believe should address this.

could you give it a run?

```yaml
deploy:
- provider: npm
  edge: 
    branch: sf-run_script
```

kripod:
It [works flawlessly](https://www.travis-ci.com/kripod/react-hooks/jobs/249181556), thank you once again!

PR #1124, opened 2019-10-17
add npm --run_script

svenfuchs:
see https://github.com/travis-ci/dpl/issues/1118

allows one or many scripts to be run using `npm run [script]`. e.g.

```yaml
deploy:
- provider: npm
  edge: 
    branch: sf-run_script
  run_script: publish
```

or

```yaml
deploy:
- provider: npm
  edge: 
    branch: sf-run_script
  run_script: 
  - one
  - other
```

### DIFF
--- a/lib/dpl/providers/npm.rb
+++ b/lib/dpl/providers/npm.rb
@@ -19,7 +19,7 @@ module Dpl
       opt '--registry URL', 'npm registry url'
       opt '--src SRC', 'directory or tarball to publish', default: '.'
       opt '--tag TAGS', 'distribution tags to add'
-      opt '--run_script', 'whether to run the publish script from package.json'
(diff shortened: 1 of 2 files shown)


lib/dpl/providers/npm.rb
```diff
@@ -42,7 +45,7 @@ def login
       end
 
       def deploy
-        shell :deploy
+        shell run_script? ? :run : :publish
```
BanzaiMan: I am not a big fan of this construct. It caught me by surprise how difficult it is to reason. Can we do a more straightforward

```ruby
if run_script?
  shell :run
else
  shell :publish
end
```

svenfuchs: sure thing!

Issue #985, opened 2019-04-12
elasticbeanstalk deployment with wait_until_deployed succeeds too soon


kzidane:
https://travis-ci.com/cs50/mailman-eb/jobs/192588278#L485 once "Too many failures" is encountered the build seems to succeed immediately before the deployment actually succeeds.

kzidane:
Hey @BanzaiMan! Was just wondering if you got a chance to look at this one?

svenfuchs:
@kzidane does this issue still exist? i'm failing to find that repository ...

kzidane:
@svenfuchs ah sorry, we renamed the repo a while ago. Here https://travis-ci.com/cs50/mailman/jobs/192588278#L483!

cloudhary:
I'm facing a similar issue. My Beanstalk deployments take a longer time than the one that is allocated in the the current deploy code: https://github.com/travis-ci/dpl/blame/f21be87991a3007480a26f3df2bbecd7fddea153/lib/dpl/providers/elasticbeanstalk.rb#L115-L123

@svenfuchs would you be willing to make the timeout longer, or allow for the user to specify a longer timeout before calling "Too many failures"?

cloudhary:
I'd be happy to make a PR to address this. We're currently debating using `eb-cli` instead in order to make our deploys wait for a success or failure from beanstalk. Fixing this by allowing user to specify timeout and using your existing codebase would be a far superior option than our own, handrolled eb-cli in the travis config. Let me know!

svenfuchs:
@cloudhary absolutely, i'd be happy to accept your PR for this. thanks in advance!

svenfuchs:
should be fixed in https://github.com/travis-ci/dpl/pull/1133

kzidane:
@svenfuchs fyi I'm afraid this still seems to be an issue:
https://travis-ci.com/cs50/cs50.me/builds/136826707
https://travis-ci.com/cs50/cs50.me/builds/136827034

PR #1133, opened 2019-11-01
Specify time to wait for deploy on AWS Elastic Beanstalk

cloudhary:
Currently, when deploying to Elastic Beanstalk with the `wait_until_deployed` flag set, `dpl` will only wait for up to 100 seconds before declaring the build has failed with a "Too many failures" error. However, this is not indicative of what the actual deploy status is. 

I raised this to @svenfuchs in https://github.com/travis-ci/dpl/issues/985. EB deploys generally take much more than 100 seconds when you choose rolling or immutable deploys. Mine take about 10 mins. In line with that, this PR:

- allows user to specify a time period to wait with a new `wait_until_deployed_timeout` flag
- use a default timeout of 600 seconds to allow for more complex deployments to succeed before terminating build
- prompt the user to use the new `wait_until_deployed_timeout` flag when the build terminates before application is in Ready state

Tests:
- Failing case of a timeout that is too short for the deploy to complete here: https://travis-ci.com/cloudhary/test-travis-eb-deploy/builds/134615440
- Success case of a deploy that takes about 300 seconds here: https://travis-ci.com/cloudhary/test-travis-eb-deploy/builds/134608445

### DIFF
--- a/lib/dpl/providers/elasticbeanstalk.rb
+++ b/lib/dpl/providers/elasticbeanstalk.rb
@@ -123,12 +123,12 @@ module Dpl
 
       def wait_until_deployed
         msgs = []
-        1.upto(20) { return if check_deployment(msgs) }
+        1.upto(wait_until_deployed_timeout / 5) { return if check_deployment(msgs) }
         error 'Deploy status unknown due to timeout. Increase the wait_until_deployed_timeout option'
       end
 
       def check_deployment(msgs)
-        sleep wait_until_deployed_timeout / 20
+        sleep 5
         events.each do |event|
           msg = "#{event.event_date} [#{event.severity}] #{event.message}"
           error "Deployment failed: #{msg}" if event.severity == 'ERROR'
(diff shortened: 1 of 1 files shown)


lib/dpl/providers/elasticbeanstalk.rb
```diff
@@ -123,11 +124,11 @@ def update_app
       def wait_until_deployed
         msgs = []
         1.upto(20) { return if check_deployment(msgs) }
-        error 'Too many failures'
+        error 'Deploy status unknown due to timeout. Increase the wait_until_deployed_timeout option'
       end
 
       def check_deployment(msgs)
-        sleep 5
+        sleep wait_until_deployed_timeout / 20
```
svenfuchs: this looks great @cloudhary. my only nitpick would be that with the default 600 `wait_until_deployed_timeout / 20` would wait for 30 seconds. should we maybe check a little more often?

cloudhary: That's a solid point, didn't think of it. I've just made the change. I'll go back and test it with the builds I have

cloudhary: Both test builds above still behave as intended with this latest change. Thanks @svenfuchs for catching this!

PR #1134, opened 2019-11-11
npm: Sanitize shell input

native-api:
https://travis-ci.community/t/cannot-publish-npm-package-env-to-no-such-file-or-directory/5880

### DIFF
--- a/lib/dpl/provider/npm.rb
+++ b/lib/dpl/provider/npm.rb
@@ -1,4 +1,5 @@
 require 'json'
+require 'shellwords'
 require 'uri'
 
 module DPL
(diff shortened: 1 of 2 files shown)


native-api:
Failure unrelated, caused by the newest `rubyzip` requiring Ruby 2.4+: https://travis-ci.com/travis-ci/dpl/jobs/255057077#L3167
I would provide a fix for https://github.com/travis-ci/dpl/blob/v1/dpl-elastic_beanstalk.gemspec but instead -- do you really need to test against those old Ruby versions? Workers currently use 2.4 AFAICS.

tweirtx:
Any update on this? This bug is breaking our workflow

PR #1137, opened 2019-11-15
Issue with promote flag

radekl:


svenfuchs:
looks good 👍 

technically, that accessor should return the same boolean value, or possibly `nil`, definitely an improvement though

radekl:
Haven't finished this PR - turned on the debug flag to test, though I removed it directly from `master` right now (a232ace)

PR #1139, opened 2019-11-18
Fixed changeset name format.

radekl:
Changeset name has constraints: 
```Member must satisfy regular expression pattern: [a-zA-Z][-a-zA-Z0-9]* ```

### DIFF
--- a/lib/dpl/providers/cloudformation.rb
+++ b/lib/dpl/providers/cloudformation.rb
@@ -40,7 +40,7 @@ module Dpl
            missing_template:  'File does not exist: %{template}',
            invalid_creds:     'Invalid credentials'
 
-      strs change_set_name:   'travis-ci-build-%{build_number}-#{Time.now.strftime('%Y%m%d%H%M')}',
+      strs change_set_name:   'travis-ci-build-%{build_number}-#{Time.now.strftime(\'%Y%m%d%H%M\')}',
            change_set_desc:   'Changeset created by Travis CI job for build #%{build_number} (%{git_sha})'
 
       def login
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by radekl** [comment]
@svenfuchs please let me know if this solution is ok, or you should rather see it with `#{now.str~}` instead of pointing directly to `Time`.

**Comment by radekl** [comment]
Code has been tested and this fix solves issue.

**Comment by svenfuchs** [comment]
@radekl doesn't this make it so that the code is only ever evaluated once, when the code loads? i think technically this works, and no one will ever embed `dpl` in a long running app, but it seems a little odd to me. i'd prefer to change the method `now`. as far as i can tell that method isn't being used anywhere else in the provider class, so it's safe to change the format it returns. WDYT?

**Comment by radekl** [comment]
It doesn't matter for me tbh where it will be defined - it can be defined centrally, although anyway there should be a test checking this constraint for the name in the end.

I found one usage outside of this dpl: https://github.com/travis-ci/dpl/blob/c27153a9e6fdbbe37953588a7c5700d18ee0bc67/spec/dpl/providers/elasticbeanstalk_spec.rb#L41 . Those are tests anyway and they're mixing `Time.now` with `now` function - may want to change that one first. 

Apart of that I see no problem to set it globally.

**Comment by svenfuchs** [comment]
@radekl i was referring to this method https://github.com/travis-ci/dpl/blob/a166b51c9482c0de29e6bb42cf3489cacf7ca7b0/lib/dpl/providers/cloudformation.rb#L200 ... which i would suggest to change like this:

```
         def now
-          Time.now.strftime('%Y-%m-%dT%H:%M:%S')
+          Time.now.strftime('%Y%m%d%H%M')
         end
```

radekl:
facepalm

Well I haven't noticed this function here. Sure it is the best place to fix it :)

PR #1140, opened 2019-11-25
Convox names

radekl:
This PR solves issue with: 
- `$` sign in environment variables.
- adds new `env_names` method to allow defining variables names to be passed from the build (apart of using `VAR_NAME_FOO=${VAR_NAME_FOO}` which with longer names make a lot of deduplications.


radekl:
I would say so, because if we were not to escape `$` sign, then Travis would resolve it to Environment variable value anyway and when it's eventually passed to DPL, it shouldn't be resolved anymore (otherwise one would have to escape it like sick number of times to pass `$` sign as-is not as an bash env variable).

PR #1150, opened 2020-01-06
Update octokit dependency to ~> 4.15.0

BanzaiMan:
Resolves https://travis-ci.community/t/there-was-an-error-while-trying-to-load-the-gem-octokit/6708
by using the new octokit release, which fixes dependency issue (See
https://github.com/octokit/octokit.rb/issues/1155).


BanzaiMan:
Ah, there is one already. Closing in favor of #1149.

PR #1160, opened 2020-02-06
[WIP] Add Basic Nuget Support

robwar:
This needs to be reconciled with `master` as some changes have gone in since I branched.

Likewise, it needs some testing goodness, but I’m not sure how difficult that will be to add as it is Windows-based.


robwar:
Konrad has tested this basic functionality with MyGet and given it the 👍

svenfuchs:
totally missed this one @robwar. thanks for sending this PR!

i have taken the liberty to rebase your branch to the current master, and force pushed to your branch. if you would like to continue working on this you'd want to:

```
git fetch
git reset --hard origin/rw-nuget
```

i'll have a closer look, but i think we can probably just merge this, as the provider is in `dev` status anyway.

PR #1162, opened 2020-02-07
Fix `[ -z $var` clauses without quotes

native-api:
https://travis-ci.community/t/pypi-deploy-failing-on-windows-couldnt-install-pip-setuptools-twine-or-wheel/6338/5


ericzolf:
Looks good to me (I would suggest to have all variables in the form `${VAR}` with curly brackets but that's only cosmetic and can wait). Can I test the patch in my pipeline or do I have to wait for merging or releasing?

PR #1168, opened 2020-02-18
Add git branch provider

svenfuchs:
I'm not sure how this provider should be named. ~Maybe `git_branch` would be a better choice than just `git`~ Named it `git_branch` for now.

The intention of this provider is to help with the following use case:

* Dpl allows introspection for command description, arguments, options etc.
* Docs leverages that and partially auto-generates dpl v2 docs from that

However, this is currently a manual process, and automating it means fiddling with pushing back to GitHub and creating a PR if necessary.

This provider would allow specifying a target branch and whether to create a PR (if none exists, yet). In the above use case it would be run as part of a `cron` build.

Besides the naming question, there is also the question how to prevent people from running into endless build loops with this. Maybe the provider should refuse to accept the target branch to be the same branch as the build is currently running on, or a similar condition.


svenfuchs:
initial test here: https://travis-ci.com/travis-ci/docs-travis-ci-com/builds/149476346 using https://github.com/travis-ci/docs-travis-ci-com/commit/484228e28660bd6d655bb6ab68f8a0f8bf59c1bb

svenfuchs:
integration test: https://travis-ci.com/travis-ci/dpl/builds/149484298#L202 ... seems to work as expected.

Issue #1164, opened 2020-02-16
pages provides assumes complete control over pages directory


adrianbroher:
The copy task of the Github pages provider currently copies the `local_dir` contents into the root of the checked out git repository.

However this prevents me from generating only a subdirectory of the Github pages.  This can be useful to use the root for the project homepage and placing the API inside a subdirectory, maybe even separated by branch or tag name.

I would like to see a new `target_dir` parameter, which allows to select the target directory, where the copy task should place the files into.  By default this could point to `.`, the root of the Github pages repository.

svenfuchs:
@adrianbroher i think this totally makes sense. would be happy to accept your PR!

adrianbroher:
Issue still persists.

PR #1170, opened 2020-02-23
Pages target dir

adrianbroher:
This PR tries to take a stab at #1164.

Never did anything related to ruby, so feedback is appreciated.

### DIFF
--- a/lib/dpl/providers/pages/git.rb
+++ b/lib/dpl/providers/pages/git.rb
@@ -46,10 +46,11 @@ module Dpl
              work_dir:            'Using temporary work directory %{work_dir}',
              committer_from_gh:   'The repo is configured to use committer user and email.',
              setup_dir:           'The source dir for deployment is %s',
+             target_dir:          'The target dir for deployment is %s',
              git_clone:           'Cloning the branch %{target_branch} from the remote repo',
              git_init:            'Initializing local git repo',
              git_checkout:        'Checking out orphan branch %{target_branch}',
-             copy_files:          'Copying %{src_dir} contents to %{work_dir}',
+             copy_files:          'Copying %{src_dir} contents to %{dst_dir}',
              git_config:          'Configuring git committer to be %{name} <%{email}>',
              prepare:             'Preparing to deploy %{target_branch} branch to gh-pages',
              git_push:            'Pushing to %{url}',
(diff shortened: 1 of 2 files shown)


lib/dpl/providers/pages/git.rb
```diff
@@ -46,10 +47,11 @@ class Git < Pages
              work_dir:            'Using temporary work directory %{work_dir}',
              committer_from_gh:   'The repo is configured to use committer user and email.',
              setup_dir:           'The source dir for deployment is %s',
+             target_dir:          'The target dir for deployment is %s',
              git_clone:           'Cloning the branch %{target_branch} from the remote repo',
              git_init:            'Initializing local git repo',
              git_checkout:        'Checking out orphan branch %{target_branch}',
-             copy_files:          'Copying %{src_dir} contents to %{work_dir}',
+             copy_files:          'Copying %{src_dir} contents to %{dst_dir}',
```
svenfuchs: i think this can just be `target_dir`, in which case we wouldn't need the extra method `dst_dir` below.

svenfuchs (commented):
this looks quite good to me @adrianbroher! i am wondering if the log output will be very useful if we omit the work dir from it, and i'm not sure we need the `dst_dir` method at all if it's just eqal to `target_dir`, but those are rather cosmetic questions.

have you been able to test this with an actual deployment?

you can trigger a deployment with your fork like so:

```yaml
deploy:
- provider: pages
  edge: 
    repo:  adrianbroher/dpl
    branch: pages-target-dir
    # ...
```

adrianbroher:
>  am wondering if the log output will be very useful if we omit the work dir from it

I can add another message to the `setup` method.

> and i'm not sure we need the dst_dir method at all if it's just eqal to target_dir

I actually wanted to add a guard/limitation so that `target_dir` can only be a decendant directory of `work_dir`.  However I have yet to find out how to implement this in ruby and if the accessor is the right place for this or if `opt` thing allows validators.

> have you been able to test this with an actual deployment?

No, but I created this [test repository](https://github.com/adrianbroher/dpl-pages-target-dir).

The .travis.yml configuration: https://github.com/adrianbroher/dpl-pages-target-dir/blob/master/.travis.yml
The gh-pages deployment branch (only subdir is populated via dpl, the root is manually maintained): https://github.com/adrianbroher/dpl-pages-target-dir/tree/gh-pages
The working build: https://travis-ci.com/adrianbroher/dpl-pages-target-dir/builds/150713363
The actual deploymeny: https://adrianbroher.github.io/dpl-pages-target-dir/

adrianbroher:
While looking at the log of the build it may be a good idea to call `git add -A %{dst_dir}` instead of `git add -A  .`.

PR #1172, opened 2020-02-26
Elastic Beanstalk provider improvements

alexjurkiewicz:
Note that edf8c70a breaks backwards compatibility by removing an option. I'm thinking this is OK because dpl v2 is still not the default choice. We could keep the old option around as an (undocumented?) compatibility shim if you prefer.

### DIFF
--- a/README.md
+++ b/README.md
@@ -326,7 +326,9 @@ Options:
   --bucket_path PATH                     Location within Bucket to upload app to (type: string)
   --description DESC                     Description for the application version (type: string)
   --label LABEL                          Label for the application version (type: string)
-  --zip_file PATH                        The zip file that you want to deploy (type: string)
+  --zip_file PATH                        The zip file that you want to deploy. You probably also should set cleanup to
+                                         false if you set this. If unset, a zipfile will be created from the current
+                                         directory, honoring .ebignore and .gitignore. (type: string)
   --[no-]wait_until_deployed             Wait until the deployment has finished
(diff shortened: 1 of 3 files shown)


svenfuchs:
thanks for opening this PR, @alexjurkiewicz.  

being able to skip updating the environment sounds reasonable to me, although i don't actually have any real experience with elastic beanstalk.

so just to make sure i understand this right ... so far we require the `env` option even though people might opt out of updating the `env` entirely by specifying `only_create_app_version`, in which case having to specify `env` is useless. is that correct?

btw

> You probably also should set cleanup to false

`cleanup` is now false by default in v2, so i don't think we need to point this out.

i assume that this PR is not related to the issue you reported here, right? https://github.com/travis-ci/dpl/issues/1173

alexjurkiewicz:
Hey Sven, your understanding of workflow is correct. We upload build and
use a non Travis system to deploy to environments.

I'll update the zip docs

alexjurkiewicz:
I admit to being a bit lost fixing these tests. If you could take a quick look I'd appreciate it 🙏

alexjurkiewicz:
ping

svenfuchs:
@alexjurkiewicz sorry, i haven't been able to have a closer look at this earlier.

i have now opened a new PR here https://github.com/travis-ci/dpl/pull/1182 with your changes, plus a commit that i believe should do the right thing https://github.com/travis-ci/dpl/pull/1182

could you please review it to see if it makes sense?

PR #1185, opened 2020-03-31
If application version already exists, use it

### LINKED ISSUE #1184: Elastic Beanstalk deploy can't handle pre-existing version
```
Setup deployment
Prepare deployment
Run deployment

Application Version build-git-b85361a-ver-c77df03699 already exists. (Aws::ElasticBeanstalk::Errors::InvalidParameterValue)

failed to deploy
```

If you try and specify an application name that already exists, `dpl` returns an error. It would be nice if this could be changed to ignore the error and proceed (or overwrite the existing version -- there's no difference from our perspective.)

alexjurkiewicz:
Fixes #1184

### DIFF
--- a/lib/dpl/providers/elasticbeanstalk.rb
+++ b/lib/dpl/providers/elasticbeanstalk.rb
@@ -109,16 +109,28 @@ module Dpl
       end
 
       def create_version
-        @version = eb.create_application_version(
-          application_name: app,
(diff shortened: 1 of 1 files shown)


alexjurkiewicz:
I'm really not sure about the error handling here... again, I beg for your expert assistance Sven 🙏

Issue #1186, opened 2020-04-16
CodeDeploy failure during `wait_until_deployed` does not cause `deploy` to fail


frenata:
When using the `CodeDeploy` provider and the `wait_until_deployed` option, the deploy does not fail when the returned status is `Failed`. This can happen when working with a script that times out, returning the AWS `ScriptTimedOut` error code, but probably also for a variety of other reasons.

This is unexpected behavior: when CodeDeploy reports a failure, the corresponding dpl job running that deploy should also in general fail.

The function that does not behave as expected follows:

https://github.com/travis-ci/dpl/blob/793b282ec332b39c42107f79769a6a2847267052/lib/dpl/providers/codedeploy.rb#L84-L88

I believe that a reasonable fix would be to adjust the `wait_until_deployed` function to behave more like its `OpsWorks` counterpart by calling `error` rather than `info` when the returned status is not a successful one:

https://github.com/travis-ci/dpl/blob/793b282ec332b39c42107f79769a6a2847267052/lib/dpl/providers/opsworks.rb#L71-L77

Edit: To be clear, I'm seeing this behavior with `v2`.

frenata:
I'd be happy to prepare a patch assuming the maintainers agree this is an issue that should be fixed.

BanzaiMan:
That sounds reasonable. We welcome your PR!

PR #1188, opened 2020-04-22
Fail CodeDeploy when wait_until_deployed fails

frenata:
Specifically, if the eventually returned status is not 'Succeeded', then
the deploy fails with a message indicating that waiting was terminated
and the returned status.

This change includes a test that asserts that the failure is correctly
raised in this case.

With some additional effort I could set up a script that should provoke
a non-'Succeeded' status code and prove out the integration itself,
please let me know if that's desired.

I'm definitely also open to feedback on the style of the returned
error message.

Closes: #1186

### DIFF
--- a/lib/dpl/providers/codedeploy.rb
+++ b/lib/dpl/providers/codedeploy.rb
@@ -84,8 +84,12 @@ module Dpl
       def wait_until_deployed(id)
         print :waiting_for_deploy
         status = poll(id) until %w(Succeeded Failed Stopped).include?(status)
-        error "Waiting terminated with status: #{status}" if status != 'Succeeded'
-        info :finished_deploy, status
+        case status
+        when 'Succeeded'
+          info :finished_deploy, status
+        else
+          error :finished_deploy, status
+        end
       end
 
       def poll(id)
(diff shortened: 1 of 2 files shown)


frenata:
The errors reported by Travis don't seem relevant to the proposed changes.

lib/dpl/providers/codedeploy.rb
```diff
@@ -84,6 +84,7 @@ def create_deployment
       def wait_until_deployed(id)
         print :waiting_for_deploy
         status = poll(id) until %w(Succeeded Failed Stopped).include?(status)
+        error "Waiting terminated with status: #{status}" if status != 'Succeeded'
```
BanzaiMan, line 87: The message seems rather awkward. How about simply changing how we exit this method?

```ruby
    case status
    when 'Succeeded'
      info :finished_deploy, status # or set up a new message template
    else
      error :finished_deploy, status
    end
```

frenata: Yikes, I don't know how I missed this, gotta get my GH notifications under control I guess. ;)

I'm totally fine with that, it's a nice simplification.

PR #1208, opened 2020-08-23
Silence s3 warnings

native-api:
https://travis-ci.community/t/aws-s3-deploy-failing-due-to-log-length/9691


BanzaiMan:
Thanks for the PR. I feel that this PR belongs to upstream. This issue had been reported as https://github.com/aws/aws-sdk-ruby/issues/2323. As they are still actively releasing v2.X gems, I am sure this patch (or anything like it) would be welcome.

I will add that `dpl` would use aws-sdk v3, which does not manifest this problem.

native-api:
I can make it use SDK v3. Just thought that backporting @svenfuchs ' silencing patch from `master` that's already been in production is less likely to cause unexpected problems, and you are going to retire dpl v1 anyway (any year now :-P )

BanzaiMan:
I've pointed to this patch in the upstream issue. If they are not open to taking it in, we are happy to take this on here.

ryoqun:
Cool, so this is getting fixed?

We're also stumbled on this log flooding issue like this for some time: https://travis-ci.com/github/solana-labs/solana/jobs/369181068#L924

I wonder if there is a work-around temporally.

native-api:
@BanzaiMan I've reverted @svenfuchs 's patch since it's now upstream in https://github.com/aws/aws-sdk-ruby/pull/2413 and https://rubygems.org/gems/aws-sdk/versions/2.11.596 .

You likely still want to merge this PR for the fixes for unrelated build errors.

PR #1224, opened 2020-11-16
Install prerelease versions of `dpl-*`

native-api:
Versions automatically uploaded on commits into `travis-ci/dpl/v1` are considered prereleases and as such, are not installed by `gem install` without `--pre`.

Supplemental to https://github.com/travis-ci/travis-build/pull/1972


native-api:
Not needed. With an explicit version, `--pre` is not necessary.

PR #1243, opened 2021-09-02
pages: Fix deploy keys

0az:
### pages: Fix deploy key check on GHE

Remove a hardcoded github.com in the pages provider's deploy key validation check, which breaks deployment pipelines running against GitHub Enterprise.

### Force usage of selected identity when using SSH for git

This fixes usage of read-write deploy keys with the pages deploy_key option when the Travis default private key is only authorized for read-only access.

### pages: Show more information when pushing

This makes debugging broken deployments slightly easier, as error messages are no longer suppressed.

---

Signed-off-by: afzhou <afzhou@ibm.com>

### DIFF
--- a/lib/dpl/providers/pages/git.rb
+++ b/lib/dpl/providers/pages/git.rb
@@ -58,7 +58,7 @@ module Dpl
         cmds git_clone:           'git clone --quiet --branch="%{target_branch}" --depth=1 "%{remote_url}" . > /dev/null 2>&1',
              git_init:            'git init .',
              git_checkout:        'git checkout --orphan "%{target_branch}"',
-             check_deploy_key:    'ssh -i %{key} -T git@github.com 2>&1 | grep successful > /dev/null',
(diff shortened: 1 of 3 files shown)


lib/dpl/providers/pages/git.rb
```diff
@@ -119,7 +119,7 @@ def setup_deploy_key
           mv deploy_key, path
           chmod 0600, path
           setup_git_ssh path
-          shell :check_deploy_key, key: path
+          shell :check_deploy_key, key: path, url: opts[:url]
```
GbArc, line 122: could it be github.com by default if opts[:url] is empty?

GbArc (approved):

lib/dpl/providers/pages/git.rb
```diff
@@ -155,7 +155,8 @@ def git_commit
         end
 
         def git_push
-          shell :git_push, echo: false
+          shell 'ssh-keygen -lv -f /home/travis/.dpl/deploy_key'
```
radekl, line 158: This line has broke our pipelines. We are using token authentication and the file which is checked here is not available and build breaks due to that

PR #1245, opened 2021-12-01
Update Lambda runtime

mustafa-travisci:


Montana:
Was just checking up on `--unsafe-perms` issues with a new node install where installing `strongloop -g` breaks with an infinite loop on `node-gyp` `temp dir`, but fixed with` --unsafe-perms`.

I'll have to do more tinkering in Ruby to solve this issue within `dpl`. I'd say by EoW next week a solution will have been deployed. 

Thanks,
Montana

PR #1266, opened 2022-11-28
convox: changed convox CLI url

radekl:
Convox has changed it's official URL for CLI downloads.

Official url is: http://download.convox.com/cli/linux/convox - though it lacks SSL yet, thus the best way is to use S3 hosting for now. 

This commit is required for convox dpl to start working again.


radekl:
provider is no longer broken - convox team has fixed the redirection to a proper download page.

